PR #60, opened 2015-01-11
A plugin for PHP OPcache

yucchiy:
An implementation for #34


Songmu:
Good job!

@koemu, How do you think about it? Could you review it?

PR #64, opened 2015-01-20
A plugin for elasticache

ki38sato:
note: memcached has actually used. But redis does not use.


ki38sato:
I would rewrite because I think redis is not default.

y-uuki:
LGTM! Thanks!

PR #68, opened 2015-02-18
jvm plugin -pidfile option jstat failed

ta9to:
```
pid, err := ioutil.ReadFile(*optPidFile)
// pid : 20745\n
// pid_length : 6
```
use ReadFile case, pid contain "\n".
it replace to "" -> succeeded


Songmu:
Thank you!

PR #69, opened 2015-02-24
Fix some failed unit tests

ariarijp:
stretchr/testify is changed at Feb 7, 2015
assert.Equal() method does not convert expected and actual values


Songmu:
Thank you!!!

PR #75, opened 2015-03-31
Fix linux users, disk time metrics

koemu:
- Fix login user counter.
- Merge metrics: disk io time, read/write time.


y-uuki:
Grate PR -> Merge metrics. LGTM!

PR #82, opened 2015-04-16
add graph definition for java8 metaspace

ki38sato:
add graph definition for java8 metaspace
(I want to display only either 'metaspace' or 'permanent space'. but I have no idea.)

### DIFF
--- a/mackerel-plugin-jvm/jvm.go
+++ b/mackerel-plugin-jvm/jvm.go
@@ -75,6 +75,22 @@ func mergeStat(dst, src map[string]float64) {
 	}
 }
 
+
+
+// <Java8> https://docs.oracle.com/javase/8/docs/technotes/tools/unix/jstat.html
+// # jstat -gc <vmid>
+//  S0C    S1C    S0U    S1U      EC       EU        OC         OU       MC     MU    CCSC   CCSU   YGC     YGCT    FGC    FGCT     GCT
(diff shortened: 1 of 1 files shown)


yuuki1:
Thank you for your nice patch!

> (I want to display only either 'metaspace' or 'permanent space'. but I have no idea.)

I think it would be better to branch `metaspace` or `permanent space` by command line option like `-javaversion`.
How do you think about it?

ki38sato:
I thought 2 ideas in additional.
- no output without valid value.
- add new graph definition attribute like 'zero_fill: false'.
my 2 idea requires change of 'go-mackerel-plugin'.
The idea to use '-javaversion' requires only change of mackerel-agent-plugin.

yuuki1:
I think `1. no output without valid value.` is the simplest idea. 
I try it, then merge this PR.
Thank you for your idea.

Kuchitama:
Hello guys.
I think this is good patch.
I'd like to merge this PR early.

> (I want to display only either 'metaspace' or 'permanent space'. but I have no idea.)

Yes, I'd like to display one of them.
But I need to display `metaspace`, and we can change a graph visible setting.

Therefore I think this PR should be merged early.
I don't care that  both `metaspace` and `permament` are displayed.

Songmu:
Done at #129. Thank guys for contribution and sorry for slow response.

PR #84, opened 2015-04-28
additional help for nginx SSL sites

obaratch:
If "scheme" flag is working, which I think is, it should be mentioned in the readme.  Users may end up only setting the port and give up.


yuuki1:
LGTM!

PR #90, opened 2015-05-16
Fix an error in the number of requests of Varnish

mono0x:
When using ESI, `cache_hit`, `cache_miss`, and/or `cache_hitpass` might be counted more than once by one client request.  I think that using sum of `cache_hit`, `cache_miss`, and `cache_hitpass` as the number of requests is better than using `client_req`.


yuuki1:
thanks! LGTM!

PR #96, opened 2015-06-24
redis: add port to graph names for using multiple instances on one server

xorphitus:
If multiple Redis instances run on one server, we want to watch all status of them.
By using `-multi` option, port number is added to graph names and labels.

Redis runs as single process and single thread architecture.
If a server has many cores, I think, running multiple instances is an efficient way.

### DIFF
--- a/mackerel-plugin-redis/README.md
+++ b/mackerel-plugin-redis/README.md
@@ -6,7 +6,7 @@ Redis custom metrics plugin for mackerel.io agent.
 ## Synopsis
 
 ```shell
-mackerel-plugin-redis [-host=<hostname>] [-port=<port>] [-timeout=<time>]
+mackerel-plugin-redis [-host=<hostname>] [-port=<port>] [-timeout=<time>] [-metric-key-prefix=<prefix>]
 ```
 
 ## Example of mackerel-agent.conf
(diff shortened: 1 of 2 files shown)


Songmu:
Thank you for your pull request! Sorry for my slow response. :bow:
I see your problem and I know it is general problem of macekerel-agent-plugins.

About this p-r, I think it is better to specify prefix of graphdefs explicitly like `-metric-key-prefix="redis6380"` than implicitly. Because, Redis can use unix domain socket. (mackerel-plugin-redis doen't support unix domain socket currentry..., but we would support in near the future)

I plan `-metric-key-prefix` as common option crossover the mackerel-plugins. So could you change the option from `-multi` to `-metric-key-prefix` and adjust interfaces?

xorphitus:
Thank you for your reply.
I understand your ideas! I will fix my code.

xorphitus:
I changed the option `-multi` to `-metric-key-prefix`.

As for a graph label prefix, it is given automatically.
https://github.com/xorphitus/mackerel-agent-plugins/commit/c4d38cfb3d0c0782b3f9b9dba7b22e7f7095c43b#diff-4b78c4735dfe237f16d9d8d5c5de90a3R98

Songmu:
Thank you!

Songmu:
Released as mackerel-agent-plugins v0.11.0. Please try it. Thank you for your contribution! @xorphitus

xorphitus:
Thanks. You, too!

PR #97, opened 2015-06-25
Can specify database name option when postgresql does not has a database with the same name as the user name.

azusa:
lib/pq uses username for database name when connect to database if database name does not specify.
So mackerel-plugin-postgres cannnot connect to database  if Postgresql does not has a database with the same name as the user name, As the following error.

```
$ /usr/local/bin/mackerel-plugin-postgres -password=*** -user=+++++
2015/06/25 16:04:30 ERROR <metrics.plugin.postgres> Failed to select pg_stat_database. pq: database "+++++" does not exist
2015/06/25 16:04:30 OutputValues:  pq: database "+++++" does not exist
```
I added a argument option for specfy database, then this plugin became able to connect a database.

thanks

### DIFF
--- a/mackerel-plugin-postgres/README.md
+++ b/mackerel-plugin-postgres/README.md
@@ -6,14 +6,15 @@ PostgreSQL custom metrics plugin for mackerel.io agent.
 ## Synopsis
 
 ```shell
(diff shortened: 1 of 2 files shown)


yuuki1:
Nice patch! LGTM

PR #98, opened 2015-07-01
Add mackerel-plugin-aws-cloudfront

najeira:
Add mackerel-plugin-aws-cloudfront.

This plugin get metrics of CloudFront from CloudWatch same as ELB.


yuuki1:
LGTM

PR #99, opened 2015-07-08
Release version 0.9.3

Sixeight:
- Can specify database name option when postgresql does not has a database with the same name as the user name. #97
- Add mackerel-plugin-aws-cloudfront #98


Sixeight:
miss version

PR #101, opened 2015-07-08
fix tiny typo

hiroakis:
This is a tiny fix for README.md of Redis plugin. According to following line, you write `-host` option. However you write `-hostname` in README.md. `-host` is correct description.

https://github.com/mackerelio/mackerel-agent-plugins/blob/master/mackerel-plugin-redis/redis.go#L152


yuuki1:
LGTM

PR #114, opened 2015-09-09
add mackerel-plugin-fluentd

stanaka:
This pull request add mackerel-plugin-fluentd, which monitors fluentd http://www.fluentd.org/ .

### DIFF
--- a/mackerel-plugin-fluentd/fluentd.go
+++ b/mackerel-plugin-fluentd/fluentd.go
@@ -7,7 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"strings"
+	"regexp"
 
 	mp "github.com/mackerelio/go-mackerel-plugin-helper"
 )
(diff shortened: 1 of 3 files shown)


mackerel-plugin-fluentd/fluentd.go
```diff
@@ -0,0 +1,117 @@
...
+type FluentMonitorJSON struct {
+	Plugins []FluentdPluginMetrics `json:"plugins"`
+}
+
+func (f *FluentdMetrics) ParseStats(body []byte) (map[string]interface{}, error) {
+	var j FluentMonitorJSON
+	err := json.Unmarshal(body, &j)
+	f.plugins = j.Plugins
+	for i, _ := range f.plugins {
+		f.plugins[i].PluginIDModified = strings.Replace(f.plugins[i].PluginID, ":", "_", 1)
```
Songmu: [imo] All of invalid characters should be replaced here.

Songmu:
LGTM

mackerel-plugin-fluentd/fluentd.go
stanaka: I fixed it

PR #115, opened 2015-09-09
add mackerel-plugin-docker

stanaka:
a mackerel plugin for Docker https://www.docker.com/ .

### DIFF
--- a/mackerel-plugin-docker/docker.go
+++ b/mackerel-plugin-docker/docker.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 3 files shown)


mackerel-plugin-docker/docker.go
```diff
@@ -0,0 +1,234 @@
...
+	if err == nil {
+		return true, nil
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return true, err
+}
+
+func normalizeMetricName(str string) string {
```
Songmu: You may simplify the code as following.

```
var re = regexp.MustCompile(`[^-a-zA-Z0-9_]`)
re.ReplaceAllString(str, "_")
```

And `regexp.MustCompile` should be located outside of the func.

ref. http://ukai-go-talks.appspot.com/2014/gocon.slide#9

Songmu:
LGTM!

mackerel-plugin-docker/docker.go
stanaka: I push a commit to simplify this point.

PR #118, opened 2015-09-20
[docker] resolve cgroup path in systemd environment

Songmu:


mackerel-plugin-docker/docker.go
```diff
@@ -123,6 +124,61 @@ func (m DockerPlugin) findPrefixPath() (string, error) {
...
+	if ok, err := exists(prefix + "/memory/system.slice/"); ok && err == nil {
+		return pathSlice, nil
+	}
+	if ok, err := exists(prefix + "/memory/docker/"); ok && err == nil {
+		return pathDocker, nil
+	}
+	if ok, err := exists(prefix + "/memory/lxc/"); ok && err == nil {
+		return pathLxc, nil
+	}
+	return pathUnknown, fmt.Errorf("can't resolve runtime metrics path")
```
stanaka: [nit] It may be better to show a possibility that this error may occur in the case of no container exists.

stanaka:
LGTM

PR #122, opened 2015-10-07
support coreos and amazon linux for docker plugin

stanaka:
In Amazon Linux, the path of cgroups statistics should be `/cgroup`, and some blkio metrics do not exist.
In CoreOS, some blkio metrics are also missing.

This pull request add a path candidate for Amazon Linux and error handling of missing metrics.


Songmu:
LGTM

PR #123, opened 2015-10-07
Feature/key prefix for rds plugin

stanaka:
With `-use-identifier-as-prefix`, identifier is used as key prefix.

ex: `rds.<db-instance-identifer>.Latency.ReadLatency`


Songmu:
I think `-metric-key-prefix` option is better. Which is common option name of mackerel-agent-plugins.

ref. #96

stanaka:
I've re-create a pull request #124.

PR #124, opened 2015-10-14
Add Key prefix option for AWS RDS plugin

stanaka:
With `-metric-key-prefix` option, metric key prefix can be changed. The default value of key prefix is `rds`.

And `-metric-label-prefix` can be specified for label prefix.
Without `-metric-label-prefix`, `RDS` is used for key prefix `rds` or capitalized key prefix is used for other key prefix.

### Example
- Key: `<metric-key-prefix>.Latency.ReadLatency`
- Label: `<metric-label-prefix> Latency in second`


Songmu:
LGTM

PR #128, opened 2015-10-20
use wildcard definition & normalize xen names

naokibtn:
1. Use wildcard definition
  * ex) `xentop.cpu.#.cpu`
  * metric expression changed: `xentop.cpu.cpu_${XENNAME}` -> `xentop.cpu.${XENNAME}.cpu`

2. Normalize xen names
  * replace all `.` of xen name to `_`


yuuki1:
LGTM!

PR #130, opened 2015-10-23
[nit] correspond go vet / golint

Songmu:


Songmu:
squash merged at 09b96f6

PR #141, opened 2015-11-25
Add mackerel-plugin-unicorn

linyows:
A mackerel plugin for [Unicorn](https://github.com/defunkt/unicorn).


Songmu:
Thank you! I'll review later.

Songmu:
@linyows Sorry for slow response.

I propose one point.

Could you change metric name into `busy_wokers` and `idle_workers`, and stacked them like mackerel-plugins-apache2 and mackerl-plugins-plack?

If you have hard reason to change them, it is ok to stay them.

linyows:
@Songmu Well, that's a good suggestion.

Songmu:
Thank you!

PR #145, opened 2015-12-03
Add mackerel-plugin-murmur

mikoim:
[Murmur](http://wiki.mumble.info/wiki/Running_Murmur) custom metrics plugin for mackerel.io agent.

```
$ go run mackerel-plugin-murmur/murmur.go -host 127.0.0.1
murmur.connections.con_cur	0	1449132341
murmur.connections.con_max	100	1449132341
```


Songmu:
Thank you!

Please allow that we do not include it into core package (yum and deb) for a while.

PR #153, opened 2015-12-29
Fix error message (typo)

hfm:
`len(graphdef)` should be **8**.


itchyny:
LGTM. Thank you for your contribution.
Ref: https://github.com/hfm/mackerel-agent-plugins/blame/99f7a679f39c0af7d744cef477411e78bcb51f22/mackerel-plugin-memcached/memcached_test.go#L17 https://github.com/hfm/mackerel-agent-plugins/commit/48363fd3e690bc519371167699c06f98c44ffa0a#diff-5847ef38239b0791fbf607c8dab99d0bR17

PR #154, opened 2016-01-06
When Seconds_Behind_Master is NULL, agent-plugin doesn't send the Seconds_Behind_Master metric.

norisu0313:
mackerel-plugin-mysql use https://github.com/ziutek/mymysql.
row.Int method is "Get the nn-th value and return it as int (0 if NULL). Panic if conversion is".

It can’t tell the difference between "Seconds_Behind_Master=NULL" and "Seconds_Behind_Master=0".

When Seconds_Behind_Master is NULL, agent-plugin doesn't send the Seconds_Behind_Master metric.


Songmu:
Thank you!

PR #162, opened 2016-01-07
add unicorn plugin to package

yano3:
Add unicorn plugin to deb and RPM package


Songmu:
Thank you! It's releasing mistake...

PR #169, opened 2016-01-29
Add scheme option to Elasticsearch plugin

yano3:
Add scheme option to Elasticsearch plugin. We use Elasticsearch with ssl.


Songmu:
Thank you!

PR #173, opened 2016-02-04
Add metric key and label prefix option to Elasticsearch plugin

yano3:
Add metric key and label prefix option to Elasticsearch plugin.

This enables us to get metrics from multiple Elasticsearch servers which cannot run mackerel-agent (e.g. hosted Elasticsearch services).


Songmu:
Tnank you!

PR #187, opened 2016-02-29
[docker] Use Docker stats API

stanaka:
From docker 1.5, stats API is implemented.
https://docs.docker.com/engine/reference/api/docker_remote_api/

The docker plugin detects a version of docker and use API if the version is 1.9 or above. If not, use cgroup file for fetching stats.
You can specify the fetching method by `-method` options.


Songmu:
LGTM!

PR #190, opened 2016-03-03
Add mailq plugin

hanazuki:
I would like the standard plugin set to have a command to take the mail queue metrics.


hanazuki:
I've found some bug with postfix and will fix soon.

hanazuki:
It should work now.

karupanerura:
Great! I need it :+1:

PR #191, opened 2016-03-09
Fix helper.Tempfile in mysql.go

hfm:
If `helper.Tempfile` is empty, it seems correct to set the path to the tempfile.


Songmu:
:bow:

PR #194, opened 2016-03-11
added mackerel-plugin-gearmand

karupanerura:
Gearmand custom metrics plugin for mackerel.io agent.


Songmu:
Nice plugin!
I used to use gearmand and love it!

I merge this p-r now, but give me time
to think if this plugin should be contained
in rpm/deb packages or not.

PR #197, opened 2016-03-15
support to metric-key-prefix/metric-label-prefix option for mackerel-plugin-plack

karupanerura:


Songmu:
Thank you. It's nice and reasonable p-r.

However, `-metric-key-prefix` and `-metric-label-prefix` are common option names crossover the mackerel-plugins. So could you change the option names?

karupanerura:
OK. I fix it!

PR #198, opened 2016-03-15
Time out if jps,jinfo,jstat is hanged up

tom--bo:


Songmu:
Nice work, but there are some duplicate code.
Please define common function to run commnand with timeout.

PR #201, opened 2016-03-23
use /usr/bin/mackerel-plugin-*

naokibtn:
use `/usr/bin/mackerel-plugin-*`, and put symlinks into `/usr/local/bin/mackerel-plugin-*` (old only)


Songmu:
This change doesn't work in amazon linux with following error message.

```
ファイル /usr/bin (パッケージ mackerel-agent-plugins-0.19.1-1.noarch から) は、パッケージ filesystem-2.4.30-3.8.amzn1.x86_64 からのファイルと競合しています。
```

PR #214, opened 2016-04-06
[redis] skip to calculate capacity when CONFIG command failed

Songmu:
The elasticache for redis has no `CONFIG` command!


itchyny:
LGTM

PR #220, opened 2016-04-28
[docker] use goroutine for fetching metrics via API

stanaka:
If there are many docker containers, timeout occurs while fetching metrics, because mackerel-plugin-docker fetches metrics of each container sequentially.
To shorten the time, use goroutine.

## benchmark

- 18 containers (note: following number include header line.)
```
% docker ps | wc -l
19
```
- current version takes over 30 seconds.
```
% time /usr/local/bin/mackerel-plugin-docker -tempfile=/tmp/aa -method API > /dev/null
/usr/local/bin/mackerel-plugin-docker -tempfile=/tmp/aa -method API  0.02s user 0.02s system 0% cpu 35.754 total
```
- gorouttine version takes around 6 seconds.
```
% time ./mackerel-plugin-docker -tempfile=/tmp/aa -method API > /dev/null
./mackerel-plugin-docker -tempfile=/tmp/aa -method API > /dev/null  0.01s user 0.02s system 0% cpu 5.706 total
```

### DIFF
--- a/mackerel-plugin-docker/docker.go
+++ b/mackerel-plugin-docker/docker.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/fsouza/go-dockerclient"
 	mp "github.com/mackerelio/go-mackerel-plugin-helper"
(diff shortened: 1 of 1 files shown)


Songmu:
LGTM

PR #224, opened 2016-05-10
Support collecting top N postgres relpages

mechairoi:


mackerel-plugin-postgres/postgres.go
```diff
@@ -180,15 +189,54 @@ func fetchDatabaseSize(db *sql.DB) (map[string]float64, error) {
...
+	stat := make(map[string]float64)
+
+	for rows.Next() {
+		var relname string
+		var relpages int64
+		if err := rows.Scan(&relname, &relpages); err != nil {
+			logger.Warningf("Failed to scan %s", err)
+			continue
+		}
+		stat["postgres.relpages."+p.Database+"."+relname] = float64(relpages)
```
astj: According to META, this key should be `"relpages."+p.Database+"."+relname` ?

astj: - According to META, key prefix `postgres.` is unnecessary
- All metrics provided by this plugin is about `p.Database`, so it seems meaningless.

astj:
Thank you. Reviewed.

PR #230, opened 2016-05-28
aws-ec2-ebs: Use wildcard in the graph definitions

itchyny:
This pull request changes the metric keys and graph keys of the aws-ec2-ebs plugin. The original code creates each metric per EBS volume, creates hundreds and thousands of metrics in an organization.
ref: #110


Songmu:
Hmm, it is reasonable change, but break backward compatibility around key names.

We need to announce it before release.

LGTM. Merge it!

itchyny:
Thank you!

PR #232, opened 2016-06-10
Support password authentication of Redis

hico-horiuchi:
Like this:

```shell
$ mackerel-plugion-redis -password=password
```


Songmu:
Thank you!

PR #240, opened 2016-06-27
fix to use config.json to list up plugins for packaging

stanaka:


stanaka:
Both rpm and deb seems to include newly added packages, like php-fpm.

* rpm https://travis-ci.org/mackerelio/mackerel-agent-plugins/builds/140444138#L1341
* deb https://travis-ci.org/mackerelio/mackerel-agent-plugins/builds/140444138#L2120

Songmu:
LGTM

PR #255, opened 2016-09-02
[memcached] Support -metric-key-prefix option 

astj:
There're some plugins already supporting this option(mysql, redis, ...).


Songmu:
You can implement `PluginWithPrefix` interface on this situation.

astj:
@Songmu Thanks. Changed implementation with `PluginWithPrefix` interface.

Songmu:
LGTM!

astj:
Thanks!!

PR #260, opened 2016-09-15
Don't specify default tempfile path

astj:
ref: https://github.com/mackerelio/mackerel-agent/pull/277 / https://github.com/mackerelio/go-mackerel-plugin-helper/pull/21

With above changes about `MACKEREL_PLUGIN_WORKDIR`, plugins should place their tempfile under `MACKEREL_PLUGIN_WORKDIR`.
The fastest way to achieve it is to change plugins to implement `mp.PluginWithPrefix` interface and stopped to specify default tempfile. For `mp.PluginWithPrefix` without `tempfile` parameter, `mp` will automatically decide tempfile path under `MACKEREL_PLUGIN_WORKDIR`.

I'm sorry that I don't change all of plugins, because there are SO many! (maybe about 30 plugins are left unchanged...)
If this p-r approved, I'll change rest of them.

### DIFF
--- a/mackerel-plugin-apache2/apache2.go
+++ b/mackerel-plugin-apache2/apache2.go
@@ -14,67 +14,78 @@ import (
 	"github.com/urfave/cli"
 )
 
-// metric value structure
-var graphdef = map[string](mp.Graphs){
-	"apache2.workers": mp.Graphs{
-		Label: "Apache Workers",
-		Unit:  "integer",
-		Metrics: [](mp.Metrics){
-			mp.Metrics{Name: "busy_workers", Label: "Busy Workers", Diff: false, Stacked: true},
-			mp.Metrics{Name: "idle_workers", Label: "Idle Workers", Diff: false, Stacked: true},
-		},
-	},
-	"apache2.bytes": mp.Graphs{
-		Label: "Apache Bytes",
-		Unit:  "bytes",
-		Metrics: [](mp.Metrics){
-			mp.Metrics{Name: "bytes_sent", Label: "Bytes Sent", Diff: true, Type: "uint64"},
-		},
-	},
-	"apache2.cpu": mp.Graphs{
-		Label: "Apache CPU Load",
-		Unit:  "float",
-		Metrics: [](mp.Metrics){
-			mp.Metrics{Name: "cpu_load", Label: "CPU Load", Diff: false},
-		},
-	},
-	"apache2.req": mp.Graphs{
-		Label: "Apache Requests",
-		Unit:  "integer",
-		Metrics: [](mp.Metrics){
-			mp.Metrics{Name: "requests", Label: "Requests", Diff: true, Type: "uint64"},
-		},
-	},
-	"apache2.scoreboard": mp.Graphs{
-		Label: "Apache Scoreboard",
-		Unit:  "integer",
-		Metrics: [](mp.Metrics){
-			mp.Metrics{Name: "score-_", Label: "Waiting for connection", Diff: false, Stacked: true},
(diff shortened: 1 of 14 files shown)


astj:
With this p-r, following changes are introduced:
- Default tempfile path of those plugins are not under `/tmp` anymore, but under `MACKEREL_PLUGIN_WORKDIR`.
- apache2 and cloudfront will support `-metric-key-prefix` and `-metric-label-prefix` option.
  - I mistakingly implement them, but they should work correctly and are good things, maybe :)

haya14busa:
> The fastest way to achieve it is to change plugins to implement mp.PluginWithPrefix interface 

Indeed, `PluginWithPrefix` is recommended, but you don't have to implement  `PluginWithPrefix` interface to archive this p-r goal. Just meeting `Plugin` interface is enough.

e.g.

```patch
diff --git a/mackerel-plugin-docker/docker.go b/mackerel-plugin-docker/docker.go
index 6c4b602..b1229b5 100644
--- a/mackerel-plugin-docker/docker.go
+++ b/mackerel-plugin-docker/docker.go
@@ -465,16 +465,6 @@ func main() {
 	}
 
 	helper := mp.NewMackerelPlugin(docker)
-
-	if *optTempfile != "" {
-		helper.Tempfile = *optTempfile
-	} else {
-		helper.Tempfile = fmt.Sprintf("/tmp/mackerel-plugin-docker-%s", normalizeMetricName(*optHost))
-	}
-
-	if os.Getenv("MACKEREL_AGENT_PLUGIN_META") != "" {
-		helper.OutputDefinitions()
-	} else {
-		helper.OutputValues()
-	}
+	helper.Tempfile = *optTempfile
+	helper.Run()
 }
```

In order to archive this p-r goal, I think you don't have to use `PluginWithPrefix` and just remove hard coded assignment of helper.Tempfile (`/tmp/`)..

We don't need even migratation of `if os.Getenv("MACKEREL_AGENT_PLUGIN_META") != "" {...` to `helper.Run()`.

```patch
diff --git a/mackerel-plugin-docker/docker.go b/mackerel-plugin-docker/docker.go
index 6c4b602..886df8e 100644
--- a/mackerel-plugin-docker/docker.go
+++ b/mackerel-plugin-docker/docker.go
@@ -465,12 +465,7 @@ func main() {
 	}
 
 	helper := mp.NewMackerelPlugin(docker)
-
-	if *optTempfile != "" {
-		helper.Tempfile = *optTempfile
-	} else {
-		helper.Tempfile = fmt.Sprintf("/tmp/mackerel-plugin-docker-%s", normalizeMetricName(*optHost))
-	}
+	helper.Tempfile = *optTempfile
 
 	if os.Getenv("MACKEREL_AGENT_PLUGIN_META") != "" {
 		helper.OutputDefinitions()
```

Am I missing something?
I think it's a lot easier than implementing `PluginWithPrefix` for all plugins.

astj:
According to helper's implementation:
https://github.com/mackerelio/go-mackerel-plugin-helper/blob/master/mackerel-plugin.go#L189-L194
```go
		prefix := "default"
		if p, ok := h.Plugin.(PluginWithPrefix); ok {
			prefix = p.MetricKeyPrefix()
		}
		filename := fmt.Sprintf("mackerel-plugin-%s", prefix)
```

For all plugins which doesn't implement `PluginWithPrefix`, default tempfile name is `mackerel-plugin-default`, and I thought it's unacceptable.
But over all, I thought we should fix helper's implementation rather than implement `PluginWithPrefix` for all plugin.

astj:
Thank you for reviewing! I'll close this.

PR #262, opened 2016-09-20
build with Go 1.7

astj:
Currently mackerel-agent-plugins are built with go 1.6.2, but [according to Release notes of Go 1.7](https://golang.org/doc/go1.7), they may not work correctly on macOS 10.12.
>
Go 1.7 adds support for macOS 10.12 Sierra. This support was backported to Go 1.6.3. Binaries built with versions of Go before 1.6.3 will not work correctly on Sierra.

(mackerel-agent itself uses go 1.7)


Makefile
```diff
@@ -5,7 +5,7 @@ CURRENT_VERSION = $(shell git log --merges --oneline | perl -ne 'if(m/^.+Merge p
 
 BUILD_FLAGS = -ldflags "\
 	      -s -w \
-	      -X main.Version \"$(VERSION)\" \
+	      -X \"main.Version=$(VERSION)\" \
```
haya14busa: [note] it's not related with this p-r, but it seems we don't use `main.Version`.
It was introduced by this p-r https://github.com/mackerelio/mackerel-agent-plugins/pull/6, but it might be just included by mistake...?

I guess we can remove this in this p-r or following p-r.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.6.2
+- 1.7.1
```
haya14busa: mackerel-agent is tested by 1.7,  https://github.com/mackerelio/mackerel-agent/blob/3a3d0b1a9c5d655fe21537c0115f9b4bfa092914/.travis.yml#L4
p-r: https://github.com/mackerelio/mackerel-agent/pull/266

I guess we should specify the same version, do you have a plan to change mackerel-agent line too?
Or I think it's ok to just specify 1.7.

Makefile
astj: Oh...
I removed the `-X` !

.travis.yml
astj: Though we don't have to build agent and agent-plugins with same version of go, but to simplify I specify just 1.7 at https://github.com/mackerelio/mackerel-agent-plugins/pull/262/commits/4c06a5546a4034bf04dfb069d5cfe0425686151f.

itchyny:
LGTM

astj:
Thanks!

PR #265, opened 2016-10-05
[apache2] Support -metric-key-prefix option and get rid of default Tempfile specification

astj:
Was a part of #260 

ref: https://github.com/mackerelio/go-mackerel-plugin-helper/pull/21

maybe one might want to run multiple apache instance in a host, so I implemented `PluginWithPrefix`

### DIFF
--- a/mackerel-plugin-apache2/flags.go
+++ b/mackerel-plugin-apache2/flags.go
@@ -55,7 +55,7 @@ var cliMetricKerPrefix = cli.StringFlag{
 }
 
 var cliLabelPrefix = cli.StringFlag{
-	Name:  "label-prefix, l",
+	Name:  "metric-label-prefix, l",
 	Value: "Apache",
 	Usage: "Set metric label prefix.",
 }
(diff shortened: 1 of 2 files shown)


mackerel-plugin-apache2/flags.go
```diff
@@ -42,7 +44,18 @@ var cliStatusPage = cli.StringFlag{
...
 }
+
+var cliMetricKerPrefix = cli.StringFlag{
+	Name:  "metric-key-prefix, m",
+	Value: "apache2",
+	Usage: "Set metric key prefix.",
+}
+
+var cliLabelPrefix = cli.StringFlag{
+	Name:  "label-prefix, l",
```
haya14busa: How about using `metric-label-prefix` instead to make it compatible with other plugins?
https://github.com/mackerelio/mackerel-agent-plugins/search?utf8=%E2%9C%93&q=metric-label-prefix

haya14busa:
LGTM!

astj:
Thanks!

PR #268, opened 2016-10-12
Make *some* plugins to support MACKEREL_PLUGIN_WORKDIR

astj:
ref: https://github.com/mackerelio/mackerel-agent-plugins/pull/260 / https://github.com/mackerelio/go-mackerel-plugin-helper/pull/21 / https://github.com/mackerelio/go-mackerel-plugin-helper/pull/22

With https://github.com/mackerelio/go-mackerel-plugin-helper/pull/22, all we have to modify plugins to support MACKEREL_PLUGIN_WORKDIR is simple:
- [MUST] Migrate to go-mackerel-plugin-helper if it still used go-mackerel-plugin
- [MUST] Don't specify default Tempfile name (should delegated to helper)

In this p-r, I modified 9 of plugins.
- `uptime`: since this plugin already supports `-metric-key-prefix`, so (only) this plugin implemented `PluginWithPrefix`
- `td-table-count` and `snmp`:  they still used go-mackerel-plugin, so they are migrated to go-mackerel-plugin-helper
- other 6 plugins: Just stopped specifying default tempfile

### DIFF
--- a/mackerel-plugin-snmp/snmp.go
+++ b/mackerel-plugin-snmp/snmp.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 
 	"github.com/alouca/gosnmp"
-	mp "github.com/mackerelio/go-mackerel-plugin"
+	mp "github.com/mackerelio/go-mackerel-plugin-helper"
 )
 
 // SNMPMetrics metrics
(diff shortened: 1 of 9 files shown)


Songmu:
LGTM

astj:
Thanks!

PR #269, opened 2016-10-12
[redis] deal with MACKEREL_PLUGIN_WORKDIR

astj:
ref: #268 

- Migrated to go-mackerel-plugin-helper
- Remove default tempdir
- Since `-metric-key-prefix` is already supported, implement PluginWithPrefix interface


Songmu:
LGTM

astj:
Thanks!

PR #272, opened 2016-10-19
[redis] Fix a bug to fetch no metrics of keys and expired

yoheimuta:
The plugin can't send any valid values because [assignments to each metric](https://github.com/mackerelio/mackerel-agent-plugins/commit/ee04a0e52e161ba4a559c6129123e157b702dd4f#diff-4b78c4735dfe237f16d9d8d5c5de90a3R172) are never operated after the line started 'db'.


astj:
Looks Good. Thank you.

PR #280, opened 2016-11-03
Add new nvidia-smi plugin

ksauzz:
I'd like to add new plugin for monitoring GPUs, which utilizes [nvidia-smi](https://developer.nvidia.com/nvidia-system-management-interface)

new Metrics this plugin introduces:

- GPU Utilization
- GPU Memory Utilization
- GPU Fan Speed
- GPU Temperature
- GPU Memory Usage (free/used/total)

### DIFF
--- a/mackerel-plugin-nvidia-smi/nvidia_smi.go
+++ b/mackerel-plugin-nvidia-smi/nvidia_smi.go
@@ -1,0 +1,141 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	mp "github.com/mackerelio/go-mackerel-plugin-helper"
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+var queryOptions = []string{
+	"utilization.gpu",
+	"utilization.memory",
+	"temperature.gpu",
+	"fan.speed",
+	"memory.total",
+	"memory.used",
+	"memory.free",
+}
+
+var formatOptions = []string{
+	"noheader",
+	"nounits",
+	"csv",
+}
+
+var nvidiaSmiOptions = []string{
+	fmt.Sprintf("--format=%s", strings.Join(formatOptions, ",")),
+	fmt.Sprintf("--query-gpu=%s", strings.Join(queryOptions, ",")),
+}
+
+var metricsKeyFormats = []string{
+	"gpu.util.gpu%d",
+	"memory.util.gpu%d",
+	"temperature.gpu%d",
+	"fanspeed.gpu%d",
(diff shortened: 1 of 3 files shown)


astj:
Thank you!!

I'd like to ask you the use cases of this plugin.
What environment do you intend to run this plugin?

ksauzz:
Sorry for the late. I was on vacation in the last week.
The use-case is GPGPU. In this field, one of current popular topics is Deep Learning. The systems on which this plugin is intended to run are linux boxes with a NVIDIA GPU and a CUDA toolkit (tested CUDA 7.5 only). Thus, this plugin doesn't support some systems CUDA toolkit doesn't support.
Some of GPU instances on AWS also should work with this although I've not tested them yet.

mackerel-plugin-nvidia-smi/nvidia_smi.go
```diff
@@ -0,0 +1,144 @@
...
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+var queryOptions = []string{
+	"utilization.gpu",
+	"utilization.memory",
+	"temperature.gpu",
+	"fan.speed",
```
astj: I tested this plugin on AWS EC2 g2.2xlarge and failed to obtain metrics because `nvidia-smi` could not retrieve `fan.speed`.

```
$ ./mackerel-plugin-nvidia-smi
2016/11/18 10:09:10 OutputValues:  strconv.ParseUint: parsing "[Not Supported]": invalid syntax: 0, 0, 56, [Not Supported], 4036, 0, 4036
$ nvidia-smi --format=noheader,nounits,csv --query-gpu=utilization.gpu,utilization.memory,temperature.gpu,fan.speed,memory.total,memory.used,memory.free
0, 0, 56, [Not Supported], 4036, 0, 4036
$ nvidia-smi -q | grep -i 'fan speed'
    Fan Speed                       : N/A
```

It seems `nvidia-smi` may output NaN value like `[Not Supported]`.  So, I think it's better to skip those `[Not Supported]` value, rather than dying by parsing them.

ksauzz:
Thank you for your feedback and testing. I pushed the changes per your comments, but Travis CI failed. I'll look into it later.

astj:
@ksauzz Thank you!  CI Failure is due to `golint`, and is resolved at #285 . Please merge master branch into this branch, and CI will pass, I hope.

astj:
And changes look good!

ksauzz:
Thank you for the fix. I rebased the branch to master, and pushed it. 👍

PR #285, opened 2016-11-21
Change error strings to pass current golint

astj:
ref: https://github.com/mackerelio/mackerel-agent/pull/295

Without this change `golint` fails. example: https://travis-ci.org/mackerelio/mackerel-agent-plugins/builds/177070744


stefafafan:
Good catch! Looks good to me! :shipit:

astj:
Thanks!

PR #289, opened 2016-11-30
Change directory structure convention of each plugin

Songmu:
The reason is as follows.

- To increase code reusability
- To suppress the enlargement of the `main` package, especially the `main()` function

This patch is automatically converted by the following script.

https://github.com/mackerelio/mackerel-agent-plugins/blob/packagize/tool/packagize.pl


astj:
Maybe we need to declare new contributing guideline to follow this directory structure?

astj:
And maybe we'll have to update documentation of mackerel.io? https://mackerel.io/docs/entry/advanced/go-mackerel-plugin-helper

astj:
Changes themselves seems good.

PR #293, opened 2016-12-15
[apache2] Change metric name not to end with dot

astj:
According to [spec of `/api/v0/graph-defs/create`](https://mackerel.io/api-docs/entry/host-metrics#post-graphdef), metric name ending with dot (`.`) is invalid.

# Breaking change

A metric name will be changed.
```
custom.apache2.scoreboard.score-. => custom.apache2.scoreboard.score-
```


Songmu:
LGTM

astj:
Thanks, but I forgot to change graph definition so I pushed another commit.

PR #295, opened 2016-12-24
migrate from goamz to aws-sdk-go

astj:
# background

- https://github.com/crowdmob/goamz is deprecated
  - (And https://github.com/AdRoll/goamz is suggested)
- Currently we cannot build plugins using  crowdmob/goamz
- Maybe official SDK https://github.com/aws/aws-sdk-go is enough

# Changes introducing

 - For aws-{cloudfront, ec2-cpuredit, elasticache, elasticsearch, elb, rds}, Migrate from goamz to aws-sdk-go
- aws-ses also uses goamz, but it only uses its authentication part and we can build it now, so keep it
  - And practically it uses go-ses, which uses goamz's authentication interface, therefore it's harder to migrate to aws-sdk-go
- And I removed explicitly default Tempfile definition from some plugins
  - These changes improve compatibility with `MACKEREL_PLUGIN_WORKDIR` env.

### DIFF
--- a/mackerel-plugin-aws-cloudfront/lib/aws-cloudfront.go
+++ b/mackerel-plugin-aws-cloudfront/lib/aws-cloudfront.go
@@ -7,8 +7,10 @@ import (
 	"os"
 	"time"
 
-	"github.com/crowdmob/goamz/aws"
-	"github.com/crowdmob/goamz/cloudwatch"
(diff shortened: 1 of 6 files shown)


Songmu:
LGTM

astj:
Thanks

PR #296, opened 2016-12-26
[docker] Add timeout for API request

astj:
`Timeout: 0` means it waits forever: https://github.com/fsouza/go-dockerclient/blob/master/client.go#L565-L568


itchyny:
LGTM

astj:
Thanks!

PR #301, opened 2017-01-17
Make more plugins to support MACKEREL_PLUGIN_WORKDIR

astj:
ref: #268, #269 

For these plugins except gostats, actual default tempdir doesn't change from `/tmp/{PLUGIN-FILENAME}`, by go-mackerel-plugin-helper's behavior : https://github.com/mackerelio/go-mackerel-plugin-helper/pull/22
(For mackerel-plugin-gostats, originally default tempfile was `/tmp/mackerel-plugin-gosrv`, but will be `/tmp/mackerel-plugin-gostats`).

By this p-r, all plugins which has static default tempfile become `MACKEREL_PLUGIN_WORKDIR` ready. (There're still some plugins not `MACKEREL_PLUGIN_WORKDIR` ready, which decides default tempfile dynamically)

### DIFF
--- a/mackerel-plugin-aws-ec2-ebs/lib/aws-ec2-ebs.go
+++ b/mackerel-plugin-aws-ec2-ebs/lib/aws-ec2-ebs.go
@@ -331,11 +331,7 @@ func Do() {
 	}
 
 	helper := mp.NewMackerelPlugin(ebs)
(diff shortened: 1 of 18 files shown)


itchyny:
LGTM!

astj:
Thanks!

PR #302, opened 2017-01-18
Fix the label and scale in jvm plugin

itchyny:
Due to the references, the unit of gc times are available in seconds. However, the jvm plugin

- outputs the label as msec.
- scales incorrectly to calculate the gc time rate to 60 seconds.

I fixed

- the label of the graph of GC time (msec -> sec).
- the scale of GC time percentage.

References
  http://docs.oracle.com/javase/7/docs/technotes/tools/share/jstat.html#gcutil_option
  https://docs.oracle.com/javase/8/docs/technotes/tools/unix/jstat.html


daiksy:
LGTM!

itchyny:
Thank you!

PR #315, opened 2017-03-03
add `mackerel-plugin` command

Songmu:
- Bundle the plugins to single binary (`mackerel-plugin`)
- Dispatch plugin processing with argument or symbolic link naming
  - eg.
  - `mackerel-plugin memcached` runs memcached plugin.
  - When a symbolic link named `mackerel-plugin-memcached` was linked to `mackerel-plugin`, run the memcached plugin
- There are no breaking changes.


tool/gen_mackerel_plugin.pl
```diff
@@ -0,0 +1,52 @@
...
+
+func runPlugin(plug string) error {
+	switch plug {
+${case}\tdefault:
+		return fmt.Errorf("unknown plugin: %q", plug)
+	}
+	return nil
+}
+
+var plugins = []string{
```
haya14busa: [nit] You can use `[...]string{...}` to make plugins var array, not slice.

https://golang.org/ref/spec#Composite_literals

mackerel-plugin.go
```diff
@@ -0,0 +1,76 @@
...
+
+func printHelp() {
+	fmt.Printf(`mackerel-plugin %s
+
+Usage: mackerel-plugin <plugin> [<args>]
+
+Following plugins are available:
+    %s
+
+See `+"`mackerel-plugin <plugin> -h` "+`for more information on a specific plugin
```
astj: [nit]
filename of this binary (maybe `mackerel-plugin`) will be shown in each plugin's usage on called `mackerel-plugin <plugin> -h`.
Do you have any idea to solve this? (I think it's difficult…)
```
[github.com/mackerelio/mackerel-agent-plugins/build]$ ./mackerel-agent-plugins uptime -help
Usage of ./mackerel-agent-plugins:
  -metric-key-prefix string
    	Metric key prefix (default "uptime")
  -tempfile string
    	Temp file name
[github.com/mackerelio/mackerel-agent-plugins/build]$ ls -la ./mackerel-plugin-uptime
lrwxr-xr-x  1 astj  staff  22  3  6 16:12 ./mackerel-plugin-uptime@ -> mackerel-agent-plugins
[github.com/mackerelio/mackerel-agent-plugins/build]$ ./mackerel-plugin-uptime -help
Usage of ./mackerel-plugin-uptime:
  -metric-key-prefix string
    	Metric key prefix (default "uptime")
  -tempfile string
    	Temp file name
```

tool/gen_mackerel_plugin.pl
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env perl
+use 5.014;
+use warnings;
+use utf8;
+use autodie;
+use JSON::PP;
+
+my $json = do {
+    local $/;
+    open my $fh, '<', 'packaging/config.json';
```
astj: 👍

tool/gen_mackerel_plugin.pl
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env perl
```
astj: Do you have any idea when to run this script in our development flow?  Apparently we must run it when packaging/config.json changed, but one might forget to run.  IMO we need some automated process in somewhere (ci, release process, or something).

tool/gen_mackerel_plugin.pl
```diff
@@ -0,0 +1,52 @@
...
+
+my $json = do {
+    local $/;
+    open my $fh, '<', 'packaging/config.json';
+    <$fh>
+};
+
+my @plugins = sort @{ decode_json($json)->{plugins}};
+
+sub resolve_package {
```
astj: Currently there's no guarantee that all plugin's package is named in this way.
Should we add some documentations for contributors?

tool/gen_mackerel_plugin.pl
Songmu: I added test code to Makefile to detect missing `go generate` execution as following.


```shell
test `go generate ./... && git diff | wc -l` = 0 || (echo 'please `go generate ./...` and commit them' && exit 1)
```

tool/gen_mackerel_plugin.pl
Songmu: Add t/convention.t to check it. I'll document this convention later.

tool/gen_mackerel_plugin.pl
Songmu: I want to use `strings.Join()`, so stay it Slice.

mackerel-plugin.go
Songmu: I know this problem.
However, I don't expect much to use the `mackerel-plugin` command directly, but to be used via symlink, so it is not so much a problem.

mackerel-plugin.go
```diff
@@ -0,0 +1,76 @@
...
+
+	err = runPlugin(plug)
+
+	if err != nil {
+		return exitError
+	}
+	return exitOK
+}
+
+var version, gitcommit string
```
haya14busa: correct me if i'm wrong, but `version` is not set in build as far as I understand and `gitcommit` is unused.

mackerel-plugin.go
```diff
@@ -0,0 +1,76 @@
...
+		plug = strings.TrimPrefix(base, "mackerel-plugin-")
+	} else {
+		if len(args) < 2 {
+			printHelp()
+			return exitError
+		}
+		plug = args[1]
+		if helpReg.MatchString(plug) {
+			printHelp()
+			return exitOK
```
haya14busa: [nit] I think exit code should be `exitError` here for consistency. (or we can use `2` for printing help)

mackerel-plugin.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		osargs := []string{f}
+		osargs = append(osargs, args[2:]...)
+		os.Args = osargs
```
haya14busa: [nit] `os.Args = append([]string{f}, args[2:]...)` should be enough.

PR #321, opened 2017-03-30
Cross compile by go's native cross build, not by gox

astj:
Since mackerel-agent-plugins' build process is simple (compared to mackerel-agent), maybe native `go build` is enough.


itchyny:
LGTM

astj:
Thanks!

PR #327, opened 2017-04-10
Add AWS Lambda plugin

astj:
Original: https://github.com/astj/mackerel-plugin-aws-lambda

### DIFF
--- a/mackerel-plugin-aws-lambda/lib/aws-lambda.go
+++ b/mackerel-plugin-aws-lambda/lib/aws-lambda.go
@@ -193,8 +193,14 @@ func (p LambdaPlugin) GraphDefinition() map[string]mp.Graphs {
 			Metrics: []mp.Metrics{
 				{Name: "invocations_success", Label: "Success", Stacked: true},
 				{Name: "invocations_error", Label: "Error", Stacked: true},
-				{Name: "dead_letter_errors", Label: "Dead Letter Errors", Stacked: true},
 				{Name: "invocations_throttles", Label: "Throttles", Stacked: true},
+			},
+		},
+		"dead_letters": {
+			Label: (labelPrefix + " Dead Letter"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "dead_letter_errors", Label: "Errors"},
 			},
 		},
 		"duration": {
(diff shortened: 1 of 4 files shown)


mackerel-plugin-aws-lambda/lib/aws-lambda.go
```diff
@@ -0,0 +1,248 @@
...
+
+// getLastPoint fetches a CloudWatch metric and parse
+func getLastPointFromCloudWatch(cw cloudwatchiface.CloudWatchAPI, functionName string, metric metricsGroup) (*cloudwatch.Datapoint, error) {
+	now := time.Now()
+	statsInput := make([]*string, len(metric.Metrics))
+	for i, typ := range metric.Metrics {
+		statsInput[i] = aws.String(typ.Type)
+	}
+	input := &cloudwatch.GetMetricStatisticsInput{
+		StartTime:  aws.Time(now.Add(time.Duration(180) * time.Second * -1)), // 3 min
```
syou6162: Could you add your comment why you specified `StartTime` as 3 min?

mackerel-plugin-aws-lambda/lib/aws-lambda.go
```diff
@@ -0,0 +1,248 @@
...
+// GraphDefinition of LambdaPlugin
+func (p LambdaPlugin) GraphDefinition() map[string]mp.Graphs {
+	labelPrefix := strings.Title(p.Prefix)
+
+	graphdef := map[string]mp.Graphs{
+		"invocations": {
+			Label: (labelPrefix + " Invocations"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "invocations_success", Label: "Success", Stacked: true},
```
syou6162: [may]In AWS integration of Mackerel, default is `Stacked: false`. I think it would be good if both have the same default value.

astj: Okay. 
IMO `Stacked: true` is more useful in this graph, though :-)

mackerel-plugin-aws-lambda/lib/aws-lambda.go
astj: https://github.com/mackerelio/mackerel-agent-plugins/pull/327/commits/7528ee5283b48d4b3ccf3589c3591fde9b279fb2

syou6162:
LGTM!

astj:
Thanks!

PR #328, opened 2017-04-12
Fix the second parameter of Split

ailispaw:
to get fields correctly.


ailispaw:
It seems my misunderstanding... Sorry.

ailispaw:
But I cannot understand why it couldn't get id from fields correctly...

astj:
@ailispaw Hi. Can you explain your problem in detail? (like your `docker ps` output with your private labels/names masked)

ailispaw:
I tried to reproduce it, but I couldn't.
It may depend on my environment at that time when I restarted mackerel-agent container very often.
Never mind.

PR #331, opened 2017-04-14
Add mackerel-plugin-gunicorn

kizkoh:
This plugin is inspired by mackerel-plugin-plack.

Many monitoring SaaS tool support gunicorn monitoring using statsd daemon.  But, Gunicorn send the number of all worker to statd server.  This plugin support to distinguish active worker from idle worker without statsd daemon.

### DIFF
--- a/mackerel-plugin-gunicorn/README.md
+++ b/mackerel-plugin-gunicorn/README.md
@@ -11,7 +11,7 @@ mackerel-plugin-gunicorn [-status-file=<path>]
 
 ## Requirements
 
-This plugin require that gunicorn include the following code in its `config.py`.
+This plugin requires that gunicorn include the following code in its `config.py`.
 
 ``` python
 import fcntl
--- a/mackerel-plugin-gunicorn/README.md
+++ b/mackerel-plugin-gunicorn/README.md
@@ -1,0 +1,210 @@
+mackerel-plugin-gunicorn
+=====================
+
+Gunicorn custom metrics plugin for mackerel.io agent.
+
+## Synopsis
+
+```shell
+mackerel-plugin-gunicorn [-status-file=<path>]
+```
+
+## Requirements
+
+This plugin require that gunicorn include the following code in its `config.py`.
+
+``` python
+import fcntl
+import json
+import os
+import stat
+
+SERVER_STATUS = "/dev/shm/gunicorn_status.json"
+
+
+def on_starting(servers):
+    if os.path.exists(SERVER_STATUS):
+        os.remove(SERVER_STATUS)
+    with open(SERVER_STATUS, mode="w") as f:
+        obj = {
+            "TotalAccesses": "0",
+            "IdleWorkers": "0",
+            "BusyWorkers": "0",
+            "stats": [],
+        }
+        json.dump(obj, f)
+    os.chown(SERVER_STATUS, os.getuid(), servers.cfg.gid)
+    statinfo = os.stat(SERVER_STATUS)
+    mode = statinfo.st_mode + stat.S_IWGRP
+    os.chmod(SERVER_STATUS, mode=mode)
+
+
+def post_fork(server, worker):
+    stat = {
+        "pid": worker.pid,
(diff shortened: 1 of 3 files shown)


mackerel-plugin-gunicorn/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Synopsis
+
+```shell
+mackerel-plugin-gunicorn [-status-file=<path>]
+```
+
+## Requirements
+
+This plugin require that gunicorn include the following code in its `config.py`.
```
itchyny: [nit] require -> requires

mackerel-plugin-gunicorn/lib/gunicorn.go
```diff
@@ -0,0 +1,154 @@
...
+//       "status": "_"
+//     },
+//     {
+//       "host": "localhost:8000",
+//       "method": "GET",
+//       "uri": "/",
+//       "pid": 26724,
+//       "status": "_"
+//     }
+//   ],
```
itchyny: [nit] I think this is a JSON sample, remove trailing semicolon.

mackerel-plugin-gunicorn/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+SERVER_STATUS = "/dev/shm/gunicorn_status.json"
+
+
+def on_starting(servers):
+    if os.path.exists(SERVER_STATUS):
+        os.remove(SERVER_STATUS)
+    with open(SERVER_STATUS, mode="w") as f:
+        obj = {
+            "TotalAccesses": "0",
```
itchyny: I'm curious why you don't use numbers for these three fields.

mackerel-plugin-gunicorn/README.md
```diff
@@ -0,0 +1,210 @@
...
+                for i, _ in stats:
+                    obj["stats"][i] = stat
+
+            obj["IdleWorkers"] = str(int(obj["IdleWorkers"]) + 1)
+            obj["BusyWorkers"] = str(int(obj["BusyWorkers"]) - 1)
+
+            f.seek(0)
+            f.truncate(0)
+            json.dump(obj, f)
+            f.flush()
```
itchyny: Reading data from file, update and write. Isn't there any race condition?

itchyny: Oops, flock seems to be the answer for that, sorry.

mackerel-plugin-gunicorn/README.md
```diff
@@ -0,0 +1,210 @@
...
+        fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+        try:
+            obj = json.load(f)
+
+            stats = [(i, v) for i, v in enumerate(obj["stats"]) if v["pid"] == worker.pid]
+            if len(stats) == 0:
+                worker.log.warn("not find worker.pid: %d in stats object", worker.pid)
+            else:
+                for i, _ in stats:
+                    obj["stats"][i] = stat
```
itchyny: When `obj["stats"]` has length 5 and `stats` has length 4, `obj["stats"][4]` will not be updated nor be deleted. Is this an intended behaviour?

itchyny: Oops, I was wrong... Maybe this codes intends to update the stats with the same pid, sorry.

kizkoh: Yes, it is an element that is equal to pid of worker in obj["stats"] to want to update it.  It is `stat` and its length is 1.

itchyny: When you want to update the single stat with the same pid, I think it's better if the code looks like... `index = find_index(...); if index < 0 then {append} else {assign}`. I think `for i, _ in stats` looks as if there might be multiple occurrence of stat with the same pid.

kizkoh:
This plugin requires to write complex python code to gunicorn config.
This problem should be settled in WSGI middleware. I close this pull request.

PR #338, opened 2017-05-01
[php-fpm] Implement PluginWithPrefix interfarce

astj:
ref: #250, #255, etc...

Since mackerel-plugin-php-fpm already supports `-metric-key-prefix` option, it's natural to implement PluginWithPrefix interface.


Songmu:
LGTM

astj:
Thanks!

PR #349, opened 2017-06-02
Add aws-dynamodb plugin

astj:
ref: http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/dynamo-metricscollected.html

Original implementation: https://github.com/astj/mackerel-plugin-aws-dynamodb

### DIFF
--- a/mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
+++ b/mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
@@ -229,6 +229,9 @@ var operationalMetricsGroup = []metricsGroup{
 	{CloudWatchName: "UserErrors", Metrics: []metric{
 		{MackerelName: "UserErrors.#", Type: metricsTypeSampleCount},
 	}},
+	{CloudWatchName: "ReturnedItemCount", Metrics: []metric{
+		{MackerelName: "ReturnedItemCount.#", Type: metricsTypeAverage},
+	}},
 }
 
 // FetchMetrics fetch the metrics
@@ -330,6 +333,13 @@ func (p DynamoDBPlugin) GraphDefinition() map[string]mp.Graphs {
 			Unit:  "integer",
 			Metrics: []mp.Metrics{
 				{Name: "*", Label: "%1", Stacked: true},
+			},
+		},
+		"ReturnedItemCount": {
+			Label: (p.LabelPrefix + " ReturnedItemCount"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "*", Label: "%1"},
 			},
 		},
 		"SuccessfulRequests": {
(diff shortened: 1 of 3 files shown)


mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
```diff
@@ -0,0 +1,387 @@
...
+		},
+		"ThrottledRequests": {
+			Label: (labelPrefix + " ThrottledRequests"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "*", Label: "%1", Stacked: true},
+			},
+		},
+		"SystemErrors": {
+			Label: (labelPrefix + " SystemErrors"),
```
itchyny: Labels can be human readable, there's no reason to concatenate the words like metric names... but what do you think.

mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
```diff
@@ -0,0 +1,387 @@
...
+	if consumedWriteCapacitySum, ok := stats["ConsumedWriteCapacityUnitsSum"].(float64); ok {
+		stats["ConsumedWriteCapacityUnitsNormalized"] = consumedWriteCapacitySum / 60.0
+	}
+	return stats
+}
+
+// GraphDefinition of DynamoDBPlugin
+func (p DynamoDBPlugin) GraphDefinition() map[string]mp.Graphs {
+	labelPrefix := strings.Title(p.Prefix)
+	labelPrefix = strings.Replace(labelPrefix, "-", " ", -1)
```
itchyny: `labelPrefix` is `Dynamodb` by default but I think `DynamoDB` is better

mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
```diff
@@ -0,0 +1,387 @@
...
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "*", Label: "%1", Stacked: true},
+			},
+		},
+		"UserErrors": {
+			Label: (labelPrefix + " UserErrors"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "*", Label: "%1", Stacked: true},
```
itchyny: I'm not sure why you have set stacked true here but maybe ok.

astj:
I just recalled this plugin does not support some metrics which takes GlobalSecondaryIndexName as dimension. I'll take a look...

mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
astj: For plugins which uses CloudWatch as data sources, IMO using original CloudWatch metric names is better because users easily imagine what those metrics are taken from.

itchyny: OK

itchyny:
LGTM

itchyny:
LGTM!

astj:
Thanks!

PR #351, opened 2017-06-03
Shorten build

Songmu:
Adjust Makefile not to build the same file more than once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,11 +63,13 @@ rpm: rpm-v1 rpm-v2
 
 rpm-v1:
 	make build GOOS=linux GOARCH=386
-	rpmbuild --define "_sourcedir `pwd`"  --define "_version ${VERSION}" \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -2,13 +2,19 @@ VERSION = 0.27.1
...
+
 build: deps
-	mkdir -p build
 	for i in mackerel-plugin-*; do \
-		go build  -ldflags="-s -w" -o build/$$i \
-		`pwd | sed -e "s|${GOPATH}/src/||"`/$$i; \
+	  make $(BINDIR)/$$i; \
 	done
 
 build/mackerel-plugin: deps
```
astj: Should we also change directory structure of `build/mackerel-plugin` to use `BINDIR` ?

Songmu: Thanks for your suggestion. I will consider doing it in another issue.

PR #339, opened 2017-05-01
Use SetTempfileByBasename to support MACKEREL_PLUGIN_WORKDIR

astj:
ref: #268, #269, #301

With this p-r and #338, all plugins use `MACKEREL_PLUGIN_WORKDIR` for Tempfile by default.

PR #354, opened 2017-06-07
Add sidekiq plugin

littlekbt:


mackerel-plugin-sidekiq/lib/sidekiq.go
```diff
@@ -0,0 +1,208 @@
...
+	return m, nil
+}
+
+// Do the plugin
+func Do() {
+	optHost := flag.String("host", "localhost", "Hostname")
+	optPort := flag.String("port", "6379", "Port")
+	optPassword := flag.String("password", "", "Password")
+	optDB := flag.Int("db", 0, "DB")
+	optTempfile := flag.String("tempfile", "/tmp/mackerel-plugin-sidekiq", "Temp file name")
```
fand: - Default value of Tempfile should not use `/tmp/`.
  - If you need to dynamically set the default value, use helper.SetTempfileByBasename. (ref. #339)
  - If not, you don't have to set the default. go-mackerel-plugin-helper automatically set the default value. (ref. #301)

mackerel-plugin-sidekiq/lib/sidekiq.go
```diff
@@ -0,0 +1,208 @@
...
+	"fmt"
+	"strconv"
+
+	r "github.com/go-redis/redis"
+	mp "github.com/mackerelio/go-mackerel-plugin-helper"
+)
+
+// SidekiqPlugin for fetching metrics
+type SidekiqPlugin struct {
+	Tempfile string
```
fand: `Tempfile` seems to be unused.

littlekbt:
Thanks review.
I fixed program.

fand:
@littlekbt 
LGTM, thanks!!

PR #355, opened 2017-06-07
Implemented mackerel-plugin-redash

yoheimuta:
We plan to use this plugin in production. After our test trial in production, we are sure that it's helpful for operators monitoring redash. 

- [x] Review https://github.com/mackerelio/mackerel-agent-plugins/blob/master/CONTRIBUTING.md

image
image
image

Usage example of each graph is as follows.

- Queue count (wait)
  - Used to detect if queue is stuck
- Scheduled count
  - Ease of grasping usage situation and trend. It leads to judgment of which worker to increase
- State count (failed)
  - Used to detect failure of query

### DIFF
--- a/mackerel-plugin-redash/lib/unsafe_stats.go
+++ b/mackerel-plugin-redash/lib/unsafe_stats.go
@@ -19,7 +19,7 @@ type UnsafeTaskStats struct {
 	Scheduled bool   `json:"scheduled"`
 }
 
-// UnsafeAllTaskStats represents task states
+// UnsafeAllTaskStates represents task states
 var UnsafeAllTaskStates = []string{
 	"waiting",
 	"finished",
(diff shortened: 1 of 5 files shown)


mackerel-plugin-redash/lib/unsafe_stats.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// UnsafeTaskStats represents a task stats
+type UnsafeTaskStats struct {
+	State     string `json:"state"`
+	Scheduled bool   `json:"scheduled"`
+}
+
+// UnsafeAllTaskStats represents task states
+var UnsafeAllTaskStates = []string{
```
yoheimuta: @stefafafan Thanks! I fixed 8bf3775.

stefafafan:
@yoheimuta 
Thanks for fixing the comment!
LGTM

PR #357, opened 2017-06-08
[nit] use go 1.8.3

astj:


astj:
Hmm…… We shouldn't specify patch version, okay

PR #359, opened 2017-06-17
[accesslog] add mackerel-plugin-accesslog

Songmu:
This plugin periodically accesses the accesslog and aggregates the status and latency of the response.
It supports Apache log and ltsv log.

graphs-screenshot


mackerel-plugin-accesslog/lib/accesslog.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (p *AccesslogPlugin) getReadCloser() (io.ReadCloser, bool, error) {
+	if p.noPosFile {
+		rc, err := os.Open(p.file)
+		return rc, true, err
+	}
+	posfile := p.getPos()
+	fi, err := os.Stat(posfile)
+	// don't output count metrics when the pos file doesn't exist or is too old
```
astj: The threshold 2 minutes seems too strict? and/or is it possible to normalize stats by diff between current timestamp and Modtime?
( go-mackerel-plugin 's `calcDiff` accepts 10 min at max, and normalizes by diffTime https://github.com/mackerelio/go-mackerel-plugin/blob/master/mackerel-plugin.go#L138-L149)

mackerel-plugin-accesslog/lib/accesslog.go
```diff
@@ -0,0 +1,183 @@
...
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/Songmu/axslogparser"
```
astj: axslogparser is so simple!

(I think some might want to specify the log format to parse, especially in Apache log format)

tool/starter
```diff
@@ -0,0 +1,107 @@
...
+
+// ${struct}Plugin mackerel plugin
+type ${struct}Plugin struct {
+	prefix string
+}
+
+// MetricKeyPrefix interface for PluginWithPrefix
+func (p ${struct}Plugin) MetricKeyPrefix() string {
+	if p.prefix == "" {
+		p.prefix = "$pkg_base"
```
Songmu: The `MetricKeyPrefix` should return a default string, even if not explicitly specified. Because, the Plugin types should provide default behavior independent of flags and passing flags is to change the default behavior. It is preferable for document generation.

mackerel-plugin-accesslog/lib/accesslog.go
```diff
@@ -0,0 +1,183 @@
...
+		)
+		line := s.Text()
+		if psr == nil {
+			psr, l, err = axslogparser.GuessParser(line)
+		} else {
+			l, err = psr.Parse(line)
+		}
+		if err != nil {
+			log.Println(err)
+			continue
```
Songmu: Yes.

mackerel-plugin-accesslog/lib/accesslog.go
Songmu: It is not as simple as you think. The axslogparser covers most cases. Which support following case.

- common log
- common log with vhost
- combined log
- combined log with extended fields
- used tab as separator in apache log
- ltsv accesslog

I should doument it...

Of course I am planning to be able to specify a custom parser later.

mackerel-plugin-accesslog/lib/accesslog.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (p *AccesslogPlugin) getPos() string {
+	base := p.file + ".pos.json"
+	if p.posFile != "" {
+		if filepath.IsAbs(p.posFile) {
+			return p.posFile
+		}
+		base = p.posFile
+	}
+	return filepath.Join(pluginutil.PluginWorkDir(), "mackerel-plugin-accesslog.d", base)
```
Songmu: The ".d" is added to clearly indicate that it is a directory, because `filepath.Join(pluginutil.PluginWorkDir(), "mackerel-plugin-accesslog")` conflicts with the file genareted by plugin to store the latest value (for `fetchLastValues`).

mackerel-plugin-accesslog/lib/accesslog.go
Songmu: Because log continues while reading log, the log reading time is indefinite, so it is a little difficult to normalize. I want to treat the number of lines in the accesslog as a metric simply, and drop the values above a certain fetch interval.

mackerel-plugin-accesslog/lib/accesslog.go
itchyny: Okay, I got it.

itchyny:
LGTM

PR #360, opened 2017-06-20
Support PostgreSQL 9.6

mechairoi:


mackerel-plugin-postgres/lib/postgres.go
```diff
@@ -20,7 +23,12 @@ var graphdef = map[string]mp.Graphs{
 		Unit:  "integer",
 		Metrics: []mp.Metrics{
 			{Name: "active", Label: "Active", Diff: false, Stacked: true},
-			{Name: "waiting", Label: "Waiting", Diff: false, Stacked: true},
+			{Name: "active_waiting", Label: "Active waiting", Diff: false, Stacked: true},
+			{Name: "idle", Label: "Idle", Diff: false, Stacked: true},
+			{Name: "idle_in_transaction", Label: "Idle in transaction", Diff: false, Stacked: true},
+			{Name: "idle_in_transaction_aborted_", Label: "Idle in transaction (aborted)", Diff: false, Stacked: true},
```
itchyny: Do we need the trailing underscore?

mackerel-plugin-postgres/lib/postgres.go
```diff
@@ -255,11 +325,17 @@ func (p PostgresPlugin) FetchMetrics() (map[string]interface{}, error) {
 	}
 	defer db.Close()
 
+	version, err := fetchVersion(db)
+	if err != nil {
+		logger.Warningf("FetchMetrics: %s", err)
```
itchyny: [nit] fetchVErsion

mackerel-plugin-postgres/lib/postgres.go
```diff
@@ -239,6 +259,56 @@ func fetchDatabaseSize(db *sqlx.DB) (map[string]interface{}, error) {
 	}, nil
 }
 
+var versionRe = regexp.MustCompile("PostgreSQL (\\d+)\\.(\\d+)(\\.(\\d+))? ")
+
+type version struct {
+	first  uint
+	second uint
+	thrird uint
+}
```
itchyny: I thought major/minor would be better at a glance but, Okay, postgresql seems to change the semantics since version 10 due to the document.

mackerel-plugin-postgres/lib/postgres.go
itchyny: Okay, I understand that it uses a simple normalization from postgresql state `idle in transaction (aborted)`. But I think we may drop the trailing underscore. (I would use simple switch case statements to reject unexpected states since nobody knows what happens in the future version)

mackerel-plugin-postgres/lib/postgres.go
itchyny: Sorry, looks good as it is.

itchyny:
LGTM

mechairoi:
😃

PR #361, opened 2017-06-20
postgres: collect dbsize only if connectable

mechairoi:
`pg_dataase_size` throws error when cannot connect the database.
https://github.com/postgres/postgres/blob/REL9_6_3/src/backend/utils/adt/dbsize.c#L92-L96


itchyny:
LGTM

mechairoi:
Thanks!

PR #363, opened 2017-06-21
postgres: add metric-key-prefix

edangelion:
Add metric-key-prefix flag to support specifying multiple databases.


astj:
Looks good, but please wait #360's completion and merge it's change.

edangelion:
I have changed codes according to #360

itchyny:
Oh, you should not push the binary...

itchyny:
LGTM!

PR #365, opened 2017-06-23
Use mackerelio/golib/logging as logger, not mackerelio/mackerel-agent/logging

astj:
ref: https://github.com/mackerelio/mackerel-agent/pull/395

Since (currently) their implementations are the same, just replacing `import` is enough.


Songmu:
LGTM

astj:
Thanks!

PR #364, opened 2017-06-23
Add mackerel plugin windows process stats

tknhs:

PR #369, opened 2017-07-04
[nit] fix `-metric-key-prefix` flag

mechairoi:
ref. https://github.com/mackerelio/mackerel-agent-plugins/pull/364#discussion_r125425167


itchyny:
LGTM

PR #370, opened 2017-07-05
fix mackerel-plugin-windows-process-stats

tknhs:
- fix the command option name
- fix typo in error message


mechairoi:
Thanks! LGTM

PR #372, opened 2017-07-14
Mysql prefix

koooge:
Hello.
I want to even up mysql prefixes `Mysql foo` and `mysql.bar`.
What do you think?

I'm sorry I've not know how to test.
So I've not checked deeply.


astj:
Thank you.

> how to test.

You can test these changes by inspecting `MySQLPlugin.GraphDefinition()`, like https://github.com/mackerelio/mackerel-agent-plugins/blob/master/mackerel-plugin-mysql/lib/mysql_test.go#L19-L26

PR #374, opened 2017-07-14
Add openldap plugin 

masahide:


astj:
Thank you for the pull-request. I'll take a look in the next week, so please wait for the review :bow:

mackerel-plugin-openldap/lib/openldap.go
```diff
@@ -0,0 +1,355 @@
...
+				{Name: "statistics_entries_monitorCounter", Label: "entries", Diff: true},
+				{Name: "statistics_referrals_monitorCounter", Label: "referrals", Diff: true},
+			},
+		},
+		"connections": {
+			Label: (labelPrefix + " connections"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "connections_total_monitorCounter", Label: "total connection", Diff: true},
+				{Name: "connections_current_monitorCounter", Label: "current connection", Diff: true},
```
astj: It looks like the value of `cn=Current,cn=Connections,cn=Monitor` is just current value, so `connections_current_monitorCounter` should be set to `Diff: false`. Or am I misunderstanding?

mackerel-plugin-openldap/lib/openldap.go
```diff
@@ -0,0 +1,355 @@
...
+				{Name: "threads_max_monitoredInfo", Label: "max", Diff: false},
+				{Name: "threads_open_monitoredInfo", Label: "open", Diff: false},
+				{Name: "threads_active_monitoredInfo", Label: "active", Diff: false},
+			},
+		},
+		"statistics": {
+			Label: (labelPrefix + " statistics"),
+			Unit:  "integer",
+			Metrics: []mp.Metrics{
+				{Name: "statistics_bytes_monitorCounter", Label: "bytes", Diff: true},
```
astj: This graph contains 4 metrics and each of them has different unit (bytes, pdu, ...).
(Generally), it's better for a graph to have metrics with same units. What do you think?

mackerel-plugin-openldap/lib/openldap.go
```diff
@@ -0,0 +1,355 @@
...
+	if m.UseTLS {
+		m.l, err = ldap.DialTLS("tcp", m.TargetHost, &tls.Config{InsecureSkipVerify: m.InsecureSkipVerify})
+	} else {
+		m.l, err = ldap.Dial("tcp", m.TargetHost)
+	}
+	if err != nil {
+		logger.Errorf("Failed to Dial %s, err: %s", m.TargetHost, err)
+		os.Exit(1)
+	}
+	err = m.l.Bind(m.BindDn, m.BindPasswd)
```
astj: With this implementation, users cannot perform unauthenticated request to ldap.
I know in many cases users restrict ldap access and require binding, but it *might* be optional. (Yes, just *might*.)
What do you think?

mackerel-plugin-openldap/lib/openldap.go
```diff
@@ -0,0 +1,355 @@
...
+		return res, err
+	}
+	searchRequest := ldap.NewSearchRequest(base, ldap.ScopeBaseObject, ldap.NeverDerefAliases, 0, 0, false, "(&(objectClass=*))", []string{"ContextCSN"}, nil)
+	sr, err := l.Search(searchRequest)
+	l.Close()
+	if err != nil {
+		logger.Errorf("Failed to ldap search. %s.", err)
+		return res, err
+	}
+	return latestCSN(sr)
```
astj: I don't know much about OpenLDAP replication, so I wonder when multiple `ContextCSN` values are obtained and how to handle them.
(If all values are valid, the worst replication_delay should be `latest(masterTime) - oldest(localTime)`?)

mackerel-plugin-openldap/lib/openldap.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return nil, err
+	}
+	mergeStat(stat, ldapStatistics)
+	ldapTotalConns, err := fetchOpenldapMetrics(m.l, "cn=Total,cn=Connections,cn=Monitor", "connections_", []string{"monitorCounter"})
+	if err != nil {
+		return nil, err
+	}
+	mergeStat(stat, ldapTotalConns)
+	ldapCurrentConns, err := fetchOpenldapMetrics(m.l, "cn=Current,cn=Connections,cn=Monitor", "connections_", []string{"monitorCounter"})
```
astj: [memo]
Those connection metrics seems to be gathered by 1 search query `cn=Connections,cn=Monitor`, but this query also corrects stats about all current connections, so this 2-query-style is better. 🙆‍♂️

mackerel-plugin-openldap/lib/openldap.go
yamasaki-masahide: Authentication may not be necessary in some cases. Change this authentication process to option.

mackerel-plugin-openldap/lib/openldap.go
masahide: Since multiple CSNs are obtained like https://github.com/openldap/openldap/blob/master/tests/scripts/test058-syncrepl-asymmetric#L1211-L1216 ,
sort them and get the newest value.
And the difference between the master and slave CSN values is used as a metric.

mackerel-plugin-openldap/lib/openldap.go
masahide: Change this to a separate graph.

mackerel-plugin-openldap/lib/openldap.go
masahide: That was right.

astj:
LGTM, merged. Thank you!

PR #375, opened 2017-07-19
[aws-dynamodb] Adjust options and graph definitions

astj:
- Remove `-tempfile` option, since this plugin does not use tempfile (= does not have metrics with `Diff: true`)
- Declare operational metrics explicitly (, not wildcard) in Graph definitions
  - inactive metrics with wildcard graph definitions will be deleted
  - ref: https://mackerel.io/api-docs/entry/host-metrics#post-graphdef


stefafafan:
The changes look good to me, just wondering about metrics that do not seem to be included in the plugin (e.g. `OnlineIndexConsumedWriteCapacity` ).

astj:
Since we don't use global secondary index in our project, I omitted metrics which needs global secondary index in #349.
If someone needs I'll implement them in another p-r.

mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go
```diff
@@ -321,35 +321,70 @@ func (p DynamoDBPlugin) GraphDefinition() map[string]mp.Graphs {
 			Label: (labelPrefix + " ThrottledRequests"),
 			Unit:  "integer",
 			Metrics: []mp.Metrics{
-				{Name: "*", Label: "%1", Stacked: true},
+				{Name: "ThrottleRequests.PutItem", Label: "PutItem", Stacked: true},
```
stefafafan: Should these be `ThrottledRequests` rather than `ThrottleRequests` (no `d`) ?

astj: Right. Fixed: https://github.com/mackerelio/mackerel-agent-plugins/pull/375/commits/1c464fd6be4b0afcac674249c69df1ead2c6598b

stefafafan: Thanks 👍

stefafafan:
Looks good to me!

astj:
Thank you!

PR #378, opened 2017-07-27
[nit] Remove executable file of mackerel-plugin-multicore

itchyny:
I found an executable is included in the git repository. Rechecked with `rm -rf ./build; git clean -fd; find . -size +300k -type f`.


itchyny:
Thanks!

PR #379, opened 2017-07-27
[nit] Reject executable in git repository

itchyny:
<s>This pull request adds a simple test to reject  executables in the git repository. If an executable is committed to the repository, `make all` will quickly fail. I'm not sure there're other better solutions.</s>

I added plugin executable pattern to .gitignore so that we don't mistakenly git add executables.


mechairoi:
LGTM!

itchyny:
Thank you!

PR #385, opened 2017-08-16
[WIP]Fix mysql extended metrics

matsuu:
Adjust metrics for Opened_tables and table_cache to correct graph.


matsuu:
replace with new PR

PR #263, opened 2016-09-21
Add evicted_nonzero and reclaimed to evictions metrics.

ando-masaki:
I want to this plugin wil be simmilar to memcached_multi_evictions.

See:
https://github.com/mhwest13/Memcached-Munin-Plugin/blob/master/memcached_multi_#L409-L441

PR #388, opened 2017-08-19
[memcached] add evicted.reclaimed and evicted.nonzero_evictions

Songmu:
resolve #263 

This change add two metrics to "evictions" graph.

- nonzero_eviction
  - Number of times an item which had an explicit expire time set had to be evicted from the LRU before it expired.
- reclaimed
  - Number of times an entry was stored using memory from an expired entry

Although one query is increased to memcached, I think there is no problem because it is once a minute.


itchyny:
LGTM!

PR #389, opened 2017-08-20
[nit] [mackerel-plugin-linux] Remove a test which is not necessary and may fall sometimes

astj:
`getProc` just executes `cat` and returns its content, so imo we don't have to test its contents.
(This test causes failure on https://travis-ci.org/mackerelio/mackerel-agent-plugins/builds/266481092,  in which environment `/proc/diskstats` does not contain `"ram0"`.)


fand:
LGTM!

astj:
thank you!

PR #394, opened 2017-08-24
[plack] Don't raise errors when parsing JSON fields failed

astj:
Main purpose is to use mackerel-plugin-plack to monitor other platforms which is only PARTIALLY compatible with Plack::Middleware::ServerStatus::Lite, but I believe this is harmless for original Plack::Middleware::ServerStatus::Lite.


itchyny:
LGTM!

astj:
Thanks!

PR #395, opened 2017-08-24
Add mackerel-plugin-json

doublemarket:
This adds a plugin that gets metrics from a JSON. I wrote most of the code during the [Mackerel Plugin Hackathon #1 Tokyo](https://mackerelio.connpass.com/event/63087/).

The plugin doesn't use [go-mackerel-plugin](https://github.com/mackerelio/go-mackerel-plugin) or [go-mackerel-plugin-helper](https://github.com/mackerelio/go-mackerel-plugin-helper). The reason is that the helpers require a graph definition for metrics while this plugin dynamically generates metrics from a JSON, which makes generating the definition hard.

However, the plugin uses the similar naming scheme to the helpers since I was originally going to create this with either of the helpers.


doublemarket:
Do I have something to do to get this merged?

astj:
Ping > @Songmu

Songmu:
Good plugin. Thank you.

After merging it, please pardon that I will make some adjustments on it, and we consider whether to include it in the distribution package.

Sorry for my slow response.

PR #397, opened 2017-08-29
[redis] Change queries metric to diff of "total_commands_processed"

edangelion:


itchyny:
Redis shows the number of queries at that one second as `instantaneous_ops_per_sec`. The mackerel-agent executes the plugins once in a minute and the metric value is useless. It's surely appropriate to take the difference of the counter value `total_commands_processed` to the value at the previous execution.

LGTM!

PR #398, opened 2017-08-29
[Redis] fix expired keys

edangelion:


mackerel-plugin-redis/lib/redis.go
```diff
@@ -233,7 +233,8 @@ func (m RedisPlugin) GraphDefinition() map[string]mp.Graphs {
 			Unit:  "integer",
 			Metrics: []mp.Metrics{
 				{Name: "keys", Label: "Keys", Diff: false},
-				{Name: "expired", Label: "Expired Keys", Diff: false},
+				{Name: "expires", Label: "Expire Time Set Keys", Diff: false},
```
itchyny: I think `Keys with expiration` or `Keys with associated expire` would be better.

itchyny: or `Keys which expires`

itchyny:
LGTM!

itchyny:
Reasoning: This metric value is calculated by the summation of  `expires=` of the Keyspace section (of redis-cli info), but the result is the total count of the keys which will expire, not keys which have been expired.

PR #401, opened 2017-09-04
Shorten builds

astj:
Currently, `make all` triggers `make build` 3 times and `make build/mackerel-plugin` 2 times.
With this p-r, `make build` will be triggered 2 times and `make build/mackerel-plugins` once.


astj:
Hmm, I misunderstood current build process. I close this.

PR #402, opened 2017-09-04
[mysql]add handler graphs

matsuu:
add `handlers` and `transaction handler` graphs to ExtendedGraphs

### DIFF
--- a/mackerel-plugin-mysql/lib/mysql.go
+++ b/mackerel-plugin-mysql/lib/mysql.go
@@ -600,6 +600,32 @@ func (m MySQLPlugin) addExtendedGraphdef(graphdef map[string]mp.Graphs) map[stri
 			{Name: "Sort_scan", Label: "Sort Scan", Diff: true, Stacked: false, Type: "uint64"},
 		},
 	}
+	graphdef["handlers"] = mp.Graphs{
+		Label: labelPrefix + " handlers",
+		Unit:  "float",
+		Metrics: []mp.Metrics{
+			{Name: "Handler_write", Label: "Handler Write", Diff: true, Stacked: true, Type: "uint64"},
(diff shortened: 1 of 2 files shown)


mackerel-plugin-mysql/lib/mysql.go
```diff
@@ -600,6 +600,32 @@ func (m MySQLPlugin) addExtendedGraphdef(graphdef map[string]mp.Graphs) map[stri
...
+			{Name: "Handler_read_prev", Label: "Handler Read Prev", Diff: true, Stacked: true, Type: "uint64"},
+			{Name: "Handler_read_rnd", Label: "Handler Read Rnd", Diff: true, Stacked: true, Type: "uint64"},
+			{Name: "Handler_read_rnd_next", Label: "Handler Read Rnd Next", Diff: true, Stacked: true, Type: "uint64"},
+		},
+	}
+	graphdef["transaction_handler"] = mp.Graphs{
+		Label: labelPrefix + " transaction handler",
+		Unit:  "float",
+		Metrics: []mp.Metrics{
+			{Name: "Handler_commit", Label: "Handler Commit", Diff: true, Stacked: false, Type: "uint64"},
```
astj: For which reason this graph is `Stacked: false` but handlers graph is `Stacked: true`?

astj:
@matsuu Thank you for the pull request!
Reviewed.  It seems good, but I wrote a question. :pray:

mackerel-plugin-mysql/lib/mysql.go
matsuu: Because `transaction handler` contains `Savepoint` and `Savepoint Rollback`. They are able to be called multiple times in a transaction so stacked count is difference with transaction count.

PR #406, opened 2017-09-15
[mysql]add MyISAM related graphs

matsuu:


mackerel-plugin-mysql/lib/mysql.go
```diff
@@ -626,6 +628,25 @@ func (m MySQLPlugin) addExtendedGraphdef(graphdef map[string]mp.Graphs) map[stri
 			{Name: "Handler_savepoint_rollback", Label: "Handler Savepoint Rollback", Diff: true, Stacked: false, Type: "uint64"},
 		},
 	}
+	graphdef["myisam_indexes"] = mp.Graphs{
+		Label: labelPrefix + " myisam Indexes",
```
astj: [nit]
Should be `MyISAM`?

mackerel-plugin-mysql/lib/mysql.go
```diff
@@ -213,6 +213,8 @@ func (m MySQLPlugin) fetchShowVariables(db mysql.Conn, stat map[string]float64)
 		if err != nil {
 			log.Fatalln("FetchExtendedMetrics (Fetch Variables): ", err)
 		}
+		stat["key_buf_bytes_used"] = stat["key_buffer_size"] - stat["Key_blocks_unused"]*stat["key_cache_block_size"]
```
astj: Those stats are useless when all databases are InnoDB.
Do you think such users should just ignore them? Or is it possible to add some options like `--disable/enable_myisam` ?

(Btw, existence of those stats guaranteed even if all databases are InnoDB? I checked some of our databases and they exist, but not sure it's guaranteed or not.)

astj:
@matsuu Thank you for the pull request. I wrote some comments.

mackerel-plugin-mysql/lib/mysql.go
matsuu: I think so. But InnoDB related graphs also have `innodb` as label.

mackerel-plugin-mysql/lib/mysql.go
matsuu: Because [MySQL 5.7 still used MyISAM storage for the system tables in the MySQL schema](https://www.percona.com/blog/2016/10/11/mysql-8-0-end-myisam/). But 8.0 is not so.
Which is better to use --disable_myisam or --enable_myisam?

mackerel-plugin-mysql/lib/mysql.go
astj: Oh.
I think there's no reason to use lower-case `innodb` as label...

mackerel-plugin-mysql/lib/mysql.go
astj: > Because MySQL 5.7 still used MyISAM storage for the system tables in the MySQL schema. But 8.0 is not so.

Okay, I understand.
This implementation generates `stat["key_buf_bytes_used"]` even if the three (key_buffer_size, Key_blocks_unused, key_cache_block_size) stats don't exist.
If they don't always exist, this implementation can generate unintended stats.

astj: > Which is better to use --disable_myisam or --enable_myisam?

Hmm......
(Whichever "disable" or "enable",) the new option interacts with `--enable_extended`, so it will be complicated.  I think it's better not to add the option.  Sorry for changing my thought. :bow:

mackerel-plugin-mysql/lib/mysql.go
matsuu: fixed in f4b3759

mackerel-plugin-mysql/lib/mysql.go
matsuu: fixed in 859040d

astj:
Thank you. LGTM!

PR #408, opened 2017-09-19
Fix redundant error by golint in redis.go

shibayu36:
TravisCI build is failed with "mackerel-plugin-redis/lib/redis.go:93:2: redundant if ...; err != nil check, just return error instead" ( https://travis-ci.org/mackerelio/mackerel-agent-plugins/jobs/277135454#L1091 ).

This is because golint's specification change (https://github.com/golang/lint/pull/319).

So I fixed redis.go to pass golint.


itchyny:
LGTM

shibayu36:
thanks!

PR #409, opened 2017-09-19
Update mackerel-plugin-dynamodb plugin

astj:
- ReadThrottleEvents metrics are missing, so added
- For ReadThrotttleEvents and WriteThrottleEvents metrics, fill 0 when values are not present


itchyny:
LGTM

astj:
Thanks!

PR #410, opened 2017-09-19
[accesslog] Feature/accesslog/customize parser

karupanerura:
This patch makes be able to specify access log format by plugin option.
It's good when axslogparser failed to detect log format.


Songmu:
Thanks!

PR #423, opened 2017-10-10
[awd-dynamodb] [incompatible] remove `.` from Metrics.Name

astj:
`mp.Metrics.Name` cannot contain `.`, but some metrics in aws-dynamodb included, so fix them.
ref: https://mackerel.io/docs/entry/advanced/custom-metrics

Previous implementation causes:
- creating graph-defs fails because of wrong metric name (try to create metric with a name like `SystemErrors.PutItem` in a graph `dynamodb.SystemErrors`)
- posting metrics succeeded, but with unexpected metric name (posted as `dynamodb.SystemErrors.SystemErrors.PutItem` metrics in `dynamodb.SystemErrors.SystemErrors` graph)

With this patch, `dynamodb.SystemErrors.PutItem` metric will be generated in `dynamodb.SystemErrors` graph, but this will be a breaking change... (Both metric name and graph name will change.)


shibayu36:
LGTM!

astj:
Thanks!

PR #424, opened 2017-10-11
[docker] Add CPU Percentage metrics

astj:
<Del>This p-r depends on https://github.com/mackerelio/go-mackerel-plugin-helper/pull/28</del>

CPU Usage stats of mackerel-plugin-docker is (diff of) CPUUsage stats from docker stats: https://godoc.org/github.com/fsouza/go-dockerclient#CPUStats

So this p-r adds `docker.cpuacct_percentage.#.{user,system}` metrics, by calculating CPU usage similar way to docker cli does: https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L168-L185


mackerel-plugin-docker/lib/docker.go
```diff
@@ -238,31 +248,41 @@ func guessMethod(docker string) (string, error) {
...
+		data, err := m.getDockerPs()
+		if err != nil {
+			return nil, err
+		}
+		lines := strings.Split(data, "\n")
+		for n, line := range lines {
+			if n == 0 {
+				continue
+			}
+			fields := regexp.MustCompile(" +").Split(line, -1)
```
itchyny: Prefer `strings.Fields`.

astj: Right.
(Those two will give different results when they parse strings starting with whitespace, like `" foo bar baz"`. But it maybe does not occur here..)

mackerel-plugin-docker/lib/docker.go
```diff
@@ -525,6 +539,19 @@ func Do() {
...
+		}
+		docker.UseCpuPercentage = true
+	} else if *optCpuFormat == "usage" {
+		docker.UseCpuPercentage = false
+	} else {
+		if docker.Method == "API" {
+			docker.UseCpuPercentage = true
+		}
+	}
+
```
itchyny: The code seems to be same as the following code, but I don't care whichever, current code is acceptable.
```golang
if *optCpuFormat == "percentage" && docker.Method == "File" {
	log.Fatalf("'--cpu-format percentage' is not supported with File method.")
}
docker.UseCpuPercentage = *optCpuFormat == "percentage"
		|| docker.Method == "API" && *optCpuFormat == ""
```

itchyny:
LGTM!

mackerel-plugin-docker/lib/docker.go
```diff
@@ -65,17 +73,20 @@ var graphdef = map[string]mp.Graphs{
 			{Name: "async", Label: "Async", Diff: true, Stacked: true, Type: "uint64"},
 		},
 	},
+	// some other fields also exist in metrics, but ther're internal intermediate data
```
itchyny: [nit] `they're` or `they are`

itchyny:
LGTM

mackerel-plugin-docker/lib/docker.go
astj: 😱

PR #425, opened 2017-10-11
[unicorn] Support metric-key-prefix

astj:
Same as other plugins.

### DIFF
--- a/mackerel-plugin-unicorn/lib/unicorn.go
+++ b/mackerel-plugin-unicorn/lib/unicorn.go
@@ -77,7 +77,7 @@ func (u UnicornPlugin) GraphDefinition() map[string]mp.Graphs {
 				{Name: "memory_workeravg", Label: "Worker Average", Diff: false, Stacked: false},
 			},
 		},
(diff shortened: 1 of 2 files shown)


mackerel-plugin-unicorn/lib/unicorn.go
```diff
@@ -75,15 +56,45 @@ func (u UnicornPlugin) FetchMetrics() (map[string]interface{}, error) {
...
+		"memory": {
+			Label: (labelPrefix + " Memory"),
+			Unit:  "bytes",
+			Metrics: []mp.Metrics{
+				{Name: "memory_workers", Label: "Workers", Diff: false, Stacked: true},
+				{Name: "memory_master", Label: "Master", Diff: false, Stacked: true},
+				{Name: "memory_workeravg", Label: "Worker Average", Diff: false, Stacked: false},
+			},
+		},
+		"unicorn.workers": {
```
syou6162: `workers` may be suitable? It is ok if you intend it.

PR #428, opened 2017-10-16
[uptime] use go-mackerel-plugin instead of using go-mackerel-plugin-helper

Songmu:
Recently go-mackerel-plugin is recommended more then go-mackerel-plugin-helper.


shibayu36:
👀

shibayu36:
LGTM

PR #429, opened 2017-10-18
[gostats] Use go-mackerel-plugin instead of go-mackerel-plugin-helper

itchyny:
We recommend go-mackerel-plugin over go-mackerel-plugin-helper.


itchyny:
Thank you.

PR #246, opened 2016-06-30
add multicore plugin

daiksy:

PR #430, opened 2017-10-18
[multicore] Refactor multicore plugin

itchyny:
This pull request refactors and fixes some issues of multicore plugin. #246

- Instead of loading the content of `/proc/stat` on memory, use io.Reader to pass the file pointer.
- Use `uint64` to represent the cpu counters instead of `float64`.
- Subtract guest from user and guest_nice from nice. Since the guest count is included in user, total included the guest count twice in the previous implementation.


mackerel-plugin-multicore/lib/multicore.go
```diff
@@ -107,43 +107,43 @@ func parseProcStat(out io.Reader) (map[string]procStats, error) {
...
-			fields := strings.Fields(line)
-			key := fields[0]
-			values := fields[1:]
-
-			// skip total cpu usage
-			if key == "cpu" {
-				continue
-			}
+		if !strings.HasPrefix(line, "cpu") {
+			continue
```
mechairoi: Why dose change `break` to `continue`?

mackerel-plugin-multicore/lib/multicore.go
```diff
@@ -120,17 +121,17 @@ func parseProcStat(out io.Reader) (map[string]procStats, error) {
...
+		counterValues, err := parseCounters(values)
 		if err != nil {
 			return nil, err
 		}
 
-		total := 0.0
-		for _, v := range floatValues {
+		var total uint64
+		for _, v := range counterValues {
 			total += v
```
mechairoi: Is it necessary to consider overflows?

mackerel-plugin-multicore/lib/multicore.go
itchyny: Fixed in 442798c.

mackerel-plugin-multicore/lib/multicore.go
itchyny: Basically it doesn't matter since we take the difference of total cpu counters. So, should we skip the `last.Total > current.Total` check? I think `current.Total - last.Total` is valid even if `total += v` overflows since the subtraction of uint64 results in the actual difference (strictly speaking, the calculation fails when the cpu counter overflows twice in the duration, but it may not occur in real situation).

mechairoi:
LGTM!

itchyny:
Thank you!

PR #433, opened 2017-10-26
[cloudfront] add -metric-key-prefix option

fujiwara:
Added `-metric-key-prefix` option to support multiple cloudfront distributions.


astj:
Thank you. LGTM.

PR #435, opened 2017-10-30
[kinesis-streams] Use Sum aggregation for Kinesis streams statistics

itchyny:
Due to the documentation (http://docs.aws.amazon.com/streams/latest/dev/monitoring-with-cloudwatch.html), some metrics returns the average value for single record. I think it's not useful and Sum statistics is more preferable.


itchyny:
Thank you!

PR #436, opened 2017-11-02
Improve docker plugin

astj:
Changed graph unit and scaled `blkio.io_serviced` to get IOPS properly.

### DIFF
--- a/mackerel-plugin-docker/lib/docker.go
+++ b/mackerel-plugin-docker/lib/docker.go
@@ -37,7 +37,7 @@ var graphdef = map[string]mp.Graphs{
 	},
 	"docker.memory.#": {
 		Label: "Docker Memory",
-		Unit:  "integer",
+		Unit:  "bytes",
 		Metrics: []mp.Metrics{
 			{Name: "cache", Label: "Cache", Diff: false, Stacked: true},
 			{Name: "rss", Label: "RSS", Diff: false, Stacked: true},
--- a/mackerel-plugin-docker/lib/docker.go
+++ b/mackerel-plugin-docker/lib/docker.go
@@ -65,7 +65,7 @@ var graphdef = map[string]mp.Graphs{
 	},
 	"docker.blkio.io_service_bytes.#": {
 		Label: "Docker BlkIO Bytes",
-		Unit:  "integer",
+		Unit:  "bytes",
 		Metrics: []mp.Metrics{
 			{Name: "read", Label: "Read", Diff: true, Stacked: true, Type: "uint64"},
 			{Name: "write", Label: "Write", Diff: true, Stacked: true, Type: "uint64"},
--- a/mackerel-plugin-docker/lib/docker.go
+++ b/mackerel-plugin-docker/lib/docker.go
@@ -57,10 +57,10 @@ var graphdef = map[string]mp.Graphs{
 		Label: "Docker BlkIO IOPS",
 		Unit:  "integer",
 		Metrics: []mp.Metrics{
-			{Name: "read", Label: "Read", Diff: true, Stacked: true, Type: "uint64"},
-			{Name: "write", Label: "Write", Diff: true, Stacked: true, Type: "uint64"},
-			{Name: "sync", Label: "Sync", Diff: true, Stacked: true, Type: "uint64"},
-			{Name: "async", Label: "Async", Diff: true, Stacked: true, Type: "uint64"},
+			{Name: "read", Label: "Read", Diff: false, Stacked: true, Type: "float64"},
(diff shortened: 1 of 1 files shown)


astj:
I found `Docker BlkIO IOPS` actually indicates IO per Minute, so now I'm thinking how to handle this... 🤔

mackerel-plugin-docker/lib/docker.go
```diff
@@ -37,7 +37,7 @@ var graphdef = map[string]mp.Graphs{
 	},
 	"docker.memory.#": {
 		Label: "Docker Memory",
-		Unit:  "integer",
+		Unit:  "bytes",
```
astj: `docker.memory.#.{cache,rss}` are taken from cgroup `memory.stat`

https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt
```
memory.stat file includes following statistics

# per-memory cgroup local status
cache		- # of bytes of page cache memory.
rss		- # of bytes of anonymous and swap cache memory (includes
		transparent hugepages).
```

mackerel-plugin-docker/lib/docker.go
```diff
@@ -55,17 +55,17 @@ var graphdef = map[string]mp.Graphs{
...
+			{Name: "read", Label: "Read", Diff: true, Stacked: true, Type: "uint64", Scale: (1.0 / 60.0)},
+			{Name: "write", Label: "Write", Diff: true, Stacked: true, Type: "uint64", Scale: (1.0 / 60.0)},
+			{Name: "sync", Label: "Sync", Diff: true, Stacked: true, Type: "uint64", Scale: (1.0 / 60.0)},
+			{Name: "async", Label: "Async", Diff: true, Stacked: true, Type: "uint64", Scale: (1.0 / 60.0)},
 		},
 	},
 	"docker.blkio.io_service_bytes.#": {
 		Label: "Docker BlkIO Bytes",
-		Unit:  "integer",
+		Unit:  "bytes",
```
astj: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt
```
- blkio.io_service_bytes
	- Number of bytes transferred to/from the disk by the group. These
	  are further divided by the type of operation - read or write, sync
	  or async. First two fields specify the major and minor number of the
	  device, third field specifies the operation type and the fourth field
	  specifies the number of bytes.
```

mackerel-plugin-docker/lib/docker.go
```diff
@@ -55,17 +55,17 @@ var graphdef = map[string]mp.Graphs{
...
 	"docker.blkio.io_serviced.#": {
 		Label: "Docker BlkIO IOPS",
-		Unit:  "integer",
+		Unit:  "iops",
 		Metrics: []mp.Metrics{
-			{Name: "read", Label: "Read", Diff: true, Stacked: true, Type: "uint64"},
-			{Name: "write", Label: "Write", Diff: true, Stacked: true, Type: "uint64"},
-			{Name: "sync", Label: "Sync", Diff: true, Stacked: true, Type: "uint64"},
-			{Name: "async", Label: "Async", Diff: true, Stacked: true, Type: "uint64"},
+			{Name: "read", Label: "Read", Diff: true, Stacked: true, Type: "uint64", Scale: (1.0 / 60.0)},
```
astj: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt
```
- blkio.io_serviced
	- Number of IOs (bio) issued to the disk by the group. These
	  are further divided by the type of operation - read or write, sync
	  or async. First two fields specify the major and minor number of the
	  device, third field specifies the operation type and the fourth field
	  specifies the number of IOs.
```

mackerel-plugin-docker/lib/docker.go
```diff
@@ -55,17 +55,17 @@ var graphdef = map[string]mp.Graphs{
 	},
```
astj: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt
```
- blkio.io_queued
	- Total number of requests queued up at any given instant for this
	  cgroup. This is further divided by the type of operation - read or
	  write, sync or async.
```

mackerel-plugin-docker/lib/docker.go
astj: btw, in many Linux distributions these stats are always zero (but in Docker for Mac non-zero).
ref: https://docs.docker.com/engine/admin/runmetrics/#memory-metrics-memorystat
```
Therefore, many distros chose to not enable it by default.
```

itchyny:
LGTM

astj:
Thanks.

PR #444, opened 2017-11-15
[nit] refactor mackerel-plugin-linux

astj:
- Use `file.Open("hoge")` instead of `exec.command("cat", "hoge")`
- Use `io.Reader` as interface between reading and parsing
  - `collectWho()` is so simple therefore it still uses String as interface
- Rename `{collect,parse}Ss()` to `{collect,parse}NetworkStat()`
  - I'd like to add `netstat` implementation to this stat in future, so renamed

### DIFF
--- a/mackerel-plugin-linux/lib/linux.go
+++ b/mackerel-plugin-linux/lib/linux.go
@@ -417,18 +417,6 @@ func parseProcVmstat(r io.Reader, p *map[string]interface{}) error {
 	return nil
 }
 
-// Getting /proc/*
-func getProc(path string) (string, error) {
-	cmd := exec.Command("cat", path)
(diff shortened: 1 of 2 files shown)


mackerel-plugin-linux/lib/linux.go
```diff
@@ -399,21 +382,20 @@ func collectProcVmstat(path string, p *map[string]interface{}) error {
...
 		return err
 	}
-	err = parseProcVmstat(data, p)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	defer file.Close()
+	return parseProcVmstat(file, p)
```
syou6162: If error occurs, returning err would be better.

```
if err != nil {
        return err
}
```

PR #460, opened 2017-12-07
add that close the response body

qt-luigi:
The client must close the response body.
See Package http Overview: https://golang.org/pkg/net/http/


qt-luigi:
This is my first mackerel commit. Thank you!

PR #461, opened 2017-12-07
add defer to closing the response body, and change position it.

qt-luigi:
a present position, If StatusCode is not StatusOK then the response body do not close.


qt-luigi:
Thank you!

PR #463, opened 2017-12-11
Redis Plugin supports custom CONFIG command

dozen:
This patch is useful when CONFIG command has renamed or disabled for failsafe or security reasons.

If `-config-command=""` is specified, the metric acquisition with the CONFIG command is skipped.


astj:
Thank you! merged.

dozen:
Thanks!

PR #469, opened 2017-12-17
[mysql] Fix some InnoDB stats

astj:
ref: #462 

During review of #462, I realized some stats parsed from `show engine innodb status` can be taken from `show status` actually.  Since parsing InnoDB status are complicated and difficult, imo it's better to obtain those stats from `show status`.

And also, some stats are parsed but not used, so I removed them, (`rows_{deleted,inserted,read,updated}`).


itchyny:
LGTM

PR #470, opened 2017-12-19
[mysql] Do not exit when fetchShowInnodbStatus failed

astj:
Currently `fetchShowInnodbStatus` returns error when no results found, but exits `show engine innodb status` failed.
With this p-r, plugin-mysql does not exit and writes a log to suggest use `-disable_innodb`.


astj:
=> #474

PR #472, opened 2017-12-20
[mysql] set Diff: true for some stats which are actually counter values

astj:
Those stats has been obtained as current value from `show engine innodb status`.
With #469, mackerel-plugin-mysql obtains those stats from `show status`, but those in `show status` are counter value and needs to be set `Diff: true`, and to be scaled `*(1/60)`.


astj:
Example. Upgraded plugin-mysql around 12:07.
image

fand:
LGTM

PR #476, opened 2017-12-25
[postgres] Add amount of xlog_location change

kizkoh:
This PR adds amount of transaction location change metric as `xlog_location.xlog_location_bytes` in postgresql plugin. When postgres is in recovery, this metric is offered with `0`.

### DIFF
--- a/mackerel-plugin-postgres/lib/postgres.go
+++ b/mackerel-plugin-postgres/lib/postgres.go
@@ -408,7 +408,7 @@ func (p PostgresPlugin) GraphDefinition() map[string]mp.Graphs {
 		},
 		"xlog_location": {
 			Label: (labelPrefix + " Amount of Transaction location change"),
-			Unit:  "integer",
+			Unit:  "bytes",
 			Metrics: []mp.Metrics{
 				{Name: "xlog_location_bytes", Label: "Amount of Transaction location change (byte)", Diff: true, Stacked: false},
 			},
--- a/mackerel-plugin-postgres/lib/postgres.go
+++ b/mackerel-plugin-postgres/lib/postgres.go
@@ -190,6 +190,42 @@ func fetchDatabaseSize(db *sqlx.DB) (map[string]interface{}, error) {
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


mackerel-plugin-postgres/lib/postgres.go
```diff
@@ -365,6 +406,13 @@ func (p PostgresPlugin) GraphDefinition() map[string]mp.Graphs {
 				{Name: "temp_bytes", Label: "Temporary file size (byte)", Diff: true, Stacked: false},
 			},
 		},
+		"xlog_location": {
+			Label: (labelPrefix + " Amount of Transaction location change"),
+			Unit:  "integer",
```
itchyny: Prefer `Unit: "bytes"`

itchyny: Note that we don't mix metrics with different units, we should choose appropriate unit for each graph so that the values of the metrics are formatted properly.

kizkoh: Ok, I understand that unit is set to bytes. But the units of other metrics are integers, not bytes. Should I adapt to your comment?

https://github.com/mackerelio/mackerel-agent-plugins/pull/476/files/06497c37ff94325d8b2703b72cb2a97d2b193ec5#diff-df6813af456003abfff9f4750fc44f6eR404

itchyny: Yes, if the value is collected in bytes.

soudai:
Support for pg_xlog_location_diff () is PostgreSQL 9.2 to 9.6.
From PostgreSQL 10, the name will change to pg_wal_lsn_diff ().

https://www.postgresql.org/docs/10/static/functions-admin.html

mackerel-plugin-postgres/lib/postgres.go
kizkoh: Fixed this by https://github.com/mackerelio/mackerel-agent-plugins/pull/476/commits/1109eeeeb5584d6d73deef85c96ac19083fe33d4.

itchyny:
LGTM!

kizkoh:
Thanks!

PR #485, opened 2018-01-12
update rpm-v2 task for building Amazon Linux 2 package

hayajo:
for Amazon Linux 2.


Makefile
```diff
@@ -76,6 +76,10 @@ rpm-v2:
 	rpmbuild --define "_sourcedir `pwd`"  --define "_version ${VERSION}" \
 	  --define "buildarch x86_64" --define "dist .el7.centos" \
 	  -bb packaging/rpm/mackerel-agent-plugins-v2.spec
+	make build/mackerel-plugin GOOS=linux GOARCH=amd64
```
astj: You don't need this line, because rpmbuild can re-use the artifact built in L75

PR #486, opened 2018-01-12
make `make build` works for some plugins which moved out from this repository

### LINKED ISSUE #484: some plugins can't build
The following plugin will cause a build error. because these plugins doesn't has "lib" directory.
(these are also not included in the "mackerel-agent-plugins_0.42.0-1_all.deb" package)

```
mackerel-plugin-aws-ec2
mackerel-plugin-aws-kinesis-firehose
mackerel-plugin-aws-rekognition
mackerel-plugin-aws-waf
mackerel-plugin-gcp-compute-engine
mackerel-plugin-gearmand
mackerel-plugin-json
```

in Makefile, plugins must has "lib" directory.

```
$(BINDIR)/mackerel-plugin-%:` mackerel-plugin-%/lib/*.go
```

astj:
Fix #484

In gnumake rules, `*.go` will resolved to proper files when ANY file exists, but will be resolved to a (non-exists) file `*.go` itself when NO files found.

To avoid this, 
  - add main.go to dependency
  - use `wildcard` function to avoid `*.go`
    - ref: https://www.gnu.org/software/make/manual/html_node/Wildcard-Function.html#Wildcard-Function
  - add `.SECONDEXPANSION` to resolve `wildcard` after `%` resolved
    - ref: https://www.gnu.org/software/make/manual/html_node/Secondary-Expansion.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ BINDIR  = build/$(GOOS)/$(GOARCH)
 
 all: lint cover testconvention rpm deb
 
-$(BINDIR)/mackerel-plugin-%: mackerel-plugin-%/lib/*.go
+.SECONDEXPANSION:
+$(BINDIR)/mackerel-plugin-%: mackerel-plugin-%/main.go $$(wildcard mackerel-plugin-%/lib/*.go)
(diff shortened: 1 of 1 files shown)


Songmu:
LGTM

astj:
Thank you!

PR #487, opened 2018-01-16
Support BSD

miwarin:
[[ ISSUE ]]
1. On BSD(NetBSD, FreeBSD, OpenBSD), 'make' is 'BSD make'.
so 'make build' will use 'BSD make', on BSD.

2. 'BSD make' is not support 'The shell Function'
see. https://www.gnu.org/software/make/manual/make.html#Shell-Function

in Makefile:
  GOOS   ?= $(shell go env GOOS)
  GOARCH ?= $(shell go env GOARCH)

this is not support BSD make.
==> BSD need to use 'gmake'.
'gmake' is 'GNU make'. 'gmake' is provided package, on BSD.

3. 'make' written directry (so used 'BSD make' on BSD)
in Makefile:
  make $(BINDIR)/$$i;
  etc..

4. 'BSD make' is not support 'Pattern rules'
see. Features of GNU make
https://www.gnu.org/software/make/manual/html_node/Features.html

in Makefile:
  $(BINDIR)/mackerel-plugin-%: mackerel-plugin-%/main.go $$(wildcard mackerel-plugin-%/lib/*.go)

'BSD make' is not support 'mackerel-plugin-%'.

[[ SOLUTION ]]
'make' replace to '$(MAKE)'

see. How the MAKE Variable Works
https://www.gnu.org/software/make/manual/make.html#MAKE-Variable

In this way, '$(MAKE)' is called from the command is used.
e.g.
  'make build' => $(MAKE) is 'make'. (for Linux, this is ok.)
  'gmake build' => $(MAKE) is 'gmake'. (This is necessary for BSD.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ $(BINDIR)/mackerel-plugin-%: mackerel-plugin-%/main.go $$(wildcard mackerel-plug
 
 build: deps
 	for i in mackerel-plugin-*; do \
-	  make $(BINDIR)/$$i; \
+	  $(MAKE) $(BINDIR)/$$i; \
 	done
 
 build/mackerel-plugin: deps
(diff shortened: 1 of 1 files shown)


Songmu:
Thanks!

PR #495, opened 2018-02-06
[aws-elasticsearch] support metric-{key,label}-prefix

astj:


mackerel-plugin-aws-elasticsearch/lib/aws-elasticsearch.go
```diff
@@ -175,6 +34,24 @@ type ESPlugin struct {
...
+// MetricKeyPrefix interface for PluginWithPrefix
+func (p ESPlugin) MetricKeyPrefix() string {
+	if p.KeyPrefix == "" {
+		return "es"
+	}
+	return p.KeyPrefix
+}
+
+// MetricLabelPrefix ...
+func (p *ESPlugin) MetricLabelPrefix() string {
```
syou6162: [nit] `func (p ESPlugin) MetricLabelPrefix()` would be better

PR #506, opened 2018-04-12
[linux] collect disk stats of NVMe devices and ignore virtual/removable devices

hayajo:
- from `/proc/diskstats` to sysfs(`/sys/block/<device>`)
- collect NVMe devices(nvme0, nvme1, ...) correctly
- exclude virtual devices and removable devices


mackerel-plugin-linux/lib/linux.go
```diff
@@ -251,45 +252,61 @@ func parseProcStat(r io.Reader, p *map[string]interface{}) error {
...
-	defer file.Close()
-	return parseProcDiskstats(file, p)
-}
 
-// parsing metrics from diskstats
-func parseProcDiskstats(r io.Reader, p *map[string]interface{}) error {
-	var elapsedData []mp.Metrics
-	var rwtimeData []mp.Metrics
+	for _, d := range devices {
+		if d.Mode()&os.ModeSymlink != os.ModeSymlink {
```
shibayu36: [question] what case is this condition true?

mackerel-plugin-linux/lib/linux.go
```diff
@@ -251,45 +252,61 @@ func parseProcStat(r io.Reader, p *map[string]interface{}) error {
 	return nil
 }
 
-// collect /proc/diskstats
-func collectProcDiskstats(path string, p *map[string]interface{}) error {
-	file, err := os.Open(path)
+// correct /sys/block/<device>/stat
```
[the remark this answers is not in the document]
shibayu36: I think `collect metrics by /sys/block/<divice>/stat` is better description.

mackerel-plugin-linux/lib/linux_test.go
```diff
@@ -167,3 +142,28 @@ pswpout 113`
 	assert.EqualValues(t, stat["pswpin"], 0)
 	assert.EqualValues(t, stat["pswpout"], 113)
 }
+
+func TestCollectDiskstats(t *testing.T) {
```
shibayu36: The method name is `collectDiskStats`, so `TestCollectDiskStats` is better than `TestCollectDiskstats`?

mackerel-plugin-linux/lib/linux_test.go
```diff
@@ -167,3 +142,28 @@ pswpout 113`
...
+	_, err := os.Stat(path)
+	if err != nil {
+		return
+	}
+	p := make(map[string]interface{})
+
+	assert.Nil(t, collectDiskStats(path, &p))
+}
+
+func TestParseDiksStat(t *testing.T) {
```
shibayu36: typo? TestParseDiskStat may be correct.

mackerel-plugin-linux/lib/linux_test.go
```diff
@@ -167,3 +142,28 @@ pswpout 113`
...
+func TestCollectDiskstats(t *testing.T) {
+	path := "/sys"
+
+	_, err := os.Stat(path)
+	if err != nil {
+		return
+	}
+	p := make(map[string]interface{})
+
+	assert.Nil(t, collectDiskStats(path, &p))
```
shibayu36: Does it work correctly when running this test on non-linux OS?

mackerel-plugin-linux/lib/linux.go
hayajo: There is no case that it becomes true, but it is checked just in case.

mackerel-plugin-linux/lib/linux_test.go
hayajo: On non-linux OS, there is no '/sys' so it will be returned early.

mackerel-plugin-linux/lib/linux.go
shibayu36: I see

mackerel-plugin-linux/lib/linux_test.go
shibayu36: ok

shibayu36:
LGTM!

PR #507, opened 2018-04-13
[aws-step-functions] mackerel-plugin-aws-step-functions

syou6162:
- https://docs.aws.amazon.com/ja_jp/step-functions/latest/dg/procedure-cw-metrics.html


syou6162:
moved to https://github.com/mackerelio/mackerel-plugin-aws-step-functions

PR #509, opened 2018-04-27
fix for collectiong ioDrive(FusionIO) diskstats

hayajo:
Fix for collectiong ioDrive(FusionIO) diskstats.

- ioDrive device name is "fioa", "fiob", ...

```
$ ls -l /sys/block/
total 0
...
lrwxrwxrwx 1 root root 0 Sep 29  2017 fioa -> ../devices/virtual/block/fioa
...
```


shibayu36:
LGTM

PR #510, opened 2018-05-07
[aws-rds]add BurstBalance metric

matsuu:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Monitoring.html

> BurstBalance
> The percent of General Purpose SSD (gp2) burst-bucket I/O credits available.
> Units: Percent


itchyny:
LGTM! Thanks.

PR #511, opened 2018-05-07
[aws-rds]fix unit for some metrics

matsuu:
DiskQueueDepth: fixed typo

> DiskQueueDepth
> The number of outstanding IOs (read/write requests) waiting to access the disk.
> Units: Count

*Throughput(aurora): fixed unit

> SelectThroughput: The average number of select queries per second.
> InsertThroughput: The average number of insert queries per second.
> UpdateThroughput: The average number of update queries per second.
> DeleteThroughput: The average number of delete queries per second.
> CommitThroughput: The average number of commit operations per second.
> DDLThroughput: The average number of DDL requests per second.
> DMLThroughput: The average number of inserts, updates, and deletes per second.

cf https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/rds-metricscollected.html


itchyny:
LGTM

PR #514, opened 2018-06-13
[aws-s3-requests] CloudWatch GetMetricStatics parameters

astj:
Adjust s3 CloudWatch GetMetricStaticsInput.
- CloudWatch metrics sometimes delays more than 3 minutes, therefore increase the duration
- S3 requests statistics are available with 1 min resolution, so set period to 60


astj:
Thank you. CI failed, but this is due to #516 so I'd like to ignore this failure.

PR #517, opened 2018-06-15
Fix typo in comments

hfm:


astj:
Nice catch, thank you!

PR #519, opened 2018-07-11
[postgres]Ignore error to support Aurora

matsuu:
Because pg_current_xlog_location() is currently not supported for Amazon Aurora


mackerel-plugin-postgres/lib/postgres.go
```diff
@@ -224,7 +224,8 @@ func fetchXlogLocation(db *sqlx.DB, version version) (map[string]interface{}, er
 			rows, err := db.Query(`SELECT pg_xlog_location_diff(pg_current_xlog_location(), '0/0')`)
 			if err != nil {
 				logger.Errorf("Failed to select pg_xlog_location_diff. %s", err)
```
astj: In that case, plugin will complain "Failed to select pg_xlog_location_diff." every minute.
How about dropping this log level from Errorf to Debugf, since default loglevel is `INFO`? https://github.com/mackerelio/golib/blob/master/logging/logger.go#L20

astj:
Thank you for pull-request.  The change itself is good, but I wrote one suggestion. Please check it.

astj:
Thank you! lgtm.

PR #520, opened 2018-07-13
add MSSQL plugin

mattn:
This add MSSQL plugin that support SQL Server and SQL Express both. Below is usage of this plugin.

```
mackerel-plugin-mssql [-instance=<SQLSERVER|SQLEXPRESS>] [-tempfile=<tempfile>]
```

MSSQL plugin

### DIFF
--- a/mackerel-plugin-mssql/README.md
+++ b/mackerel-plugin-mssql/README.md
@@ -16,3 +16,9 @@ mackerel-plugin-mssql [-metric-key-prefix=<PREFIX>] [-instance=<SQLSERVER|SQLEXP
 command = "/path/to/mackerel-plugin-mssql"
 ```
 
+## Supported Version
+
+This plugin support following version (or later) of SQL servers.
+
+* Microsoft SQL Server 2017
+* Microsoft SQL Express 2017
--- a/mackerel-plugin-mssql/README.md
+++ b/mackerel-plugin-mssql/README.md
@@ -6,7 +6,7 @@ MSSQL custom metrics plugin for mackerel.io agent.
 ## Synopsis
 
 ```shell
-mackerel-plugin-mssql [-prefix=<PREFIX>] [-instance=<SQLSERVER|SQLEXPRESS>] [-tempfile=<tempfile>]
+mackerel-plugin-mssql [-metric-key-prefix=<PREFIX>] [-instance=<SQLSERVER|SQLEXPRESS>] [-tempfile=<tempfile>]
 ```
 
 ## Example of mackerel-agent.conf
--- a/mackerel-plugin-mssql/README.md
+++ b/mackerel-plugin-mssql/README.md
@@ -22,3 +22,7 @@ This plugin support following version (or later) of SQL servers.
 
 * Microsoft SQL Server 2017
 * Microsoft SQL Express 2017
+
+## Development
+
+To update lib/wmi.go, you need to install [mattn/wmi2struct](https://github.com/mattn/wmi2struct) on Windows.
--- a/mackerel-plugin-mssql/README.md
+++ b/mackerel-plugin-mssql/README.md
@@ -1,0 +1,18 @@
+mackerel-plugin-mssql
+=====================
+
+MSSQL custom metrics plugin for mackerel.io agent.
+
+## Synopsis
+
+```shell
+mackerel-plugin-mssql [-prefix=<PREFIX>] [-instance=<SQLSERVER|SQLEXPRESS>] [-tempfile=<tempfile>]
+```
+
+## Example of mackerel-agent.conf
+
+```
+[plugin.metrics.mssql]
+command = "/path/to/mackerel-plugin-mssql"
+```
+
--- a/mackerel-plugin-mssql/lib/mssql_other.go
+++ b/mackerel-plugin-mssql/lib/mssql_other.go
@@ -1,0 +1,8 @@
+// +build !windows
+
+package mpmssql
+
+// Do the plugin
+func Do() {
+	panic("The mackerel-plugin-mssql does not work on non Windows environment, of course.")
+}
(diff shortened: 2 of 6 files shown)


mackerel-plugin-mssql/lib/mssql_windows.go
```diff
@@ -0,0 +1,183 @@
...
+
+	var plugin MSSQLPlugin
+	plugin.prefix = *optPrefix
+	plugin.instance = strings.ToUpper(*optInstance)
+	helper := mp.NewMackerelPlugin(plugin)
+
+	if *optTempfile != "" {
+		helper.Tempfile = *optTempfile
+	} else {
+		helper.Tempfile = "/tmp/mackerel-plugin-mssql"
```
astj: Now default Tempfile path is set by helper automatically, so we don't have to set this default value.
ref: https://github.com/mackerelio/go-mackerel-plugin-helper/pull/22 (This is go-mackerel-plugin*-helper* 's p-r, but this feature is also backported to go-mackerel-plugin.

====

(But this is unnecessary comment if `--tempfile` option itself is removed)

mackerel-plugin-mssql/lib/mssql_windows.go
```diff
@@ -0,0 +1,183 @@
+// +build windows
+
+package mpmssql
+
+//go:generate wmi2struct -n -p mpmssql -o wmi.go Win32_PerfRawData_MSSQLSQLEXPRESS_MSSQLSQLEXPRESSBufferManager Win32_PerfRawData_MSSQLSQLEXPRESS_MSSQLSQLEXPRESSSQLStatistics Win32_PerfRawData_MSSQLSQLEXPRESS_MSSQLSQLEXPRESSGeneralStatistics Win32_PerfRawData_MSSQLSQLEXPRESS_MSSQLSQLEXPRESSAccessMethods
```
astj: Which environment does this `go generate` require?
Apparently it needs some Windows and https://github.com/mattn/wmi2struct installed, but if there're other requirements (Windows version, SQL Server installed, …) please refer it in somewhere like README.

mackerel-plugin-mssql/lib/mssql_windows.go
```diff
@@ -0,0 +1,183 @@
...
+	if m.prefix == "" {
+		m.prefix = "mssql"
+	}
+	return m.prefix
+}
+
+func (m MSSQLPlugin) name(n string) string {
+	if m.instance == "SQLSERVER" {
+		return "Win32_PerfRawData_MSSQLSERVER_SQLServer" + n
+	} else {
```
astj: There's no validation of `--instance`.  Do you think it unnecessary?

mackerel-plugin-mssql/lib/mssql_other.go
```diff
@@ -0,0 +1,8 @@
+// +build !windows
+
+package mpmssql
+
+// Do the plugin
+func Do() {
+	panic("The mackerel-plugin-mssql does not work on non Windows environment, of course.")
```
astj: I hope this plugin will also work in other platforms IN FUTURE, since SQL Server >= 2017 works in Linux :-)

mackerel-plugin-mssql/README.md
```diff
@@ -0,0 +1,18 @@
...
+=====================
+
+MSSQL custom metrics plugin for mackerel.io agent.
+
+## Synopsis
+
+```shell
+mackerel-plugin-mssql [-prefix=<PREFIX>] [-instance=<SQLSERVER|SQLEXPRESS>] [-tempfile=<tempfile>]
+```
+
```
astj: [q]
Is there any requirements for MSSQL versions/editions?
If any, it's better to be documented here.

mackerel-plugin-mssql/lib/mssql_windows.go
```diff
@@ -0,0 +1,183 @@
...
+			},
+		},
+	}
+}
+
+// Do the plugin
+func Do() {
+	optPrefix := flag.String("metric-key-prefix", "mssql", "Metric key prefix")
+	optInstance := flag.String("instance", "SQLSERVER", "Instance name of MSSQL(SQLSERVER or SQLEXPRESS)")
+	optTempfile := flag.String("tempfile", "", "Temp file name")
```
astj: If all metrics are `diff: false`, Tempfile parameter is not necessary.
ref: https://github.com/mackerelio/mackerel-agent-plugins/pull/375

mackerel-plugin-mssql/README.md
```diff
@@ -0,0 +1,18 @@
+mackerel-plugin-mssql
+=====================
+
+MSSQL custom metrics plugin for mackerel.io agent.
+
+## Synopsis
+
+```shell
+mackerel-plugin-mssql [-prefix=<PREFIX>] [-instance=<SQLSERVER|SQLEXPRESS>] [-tempfile=<tempfile>]
```
astj: Not `-prefix` but `-metric-key-prefix` ?

mackerel-plugin-mssql/lib/mssql_windows.go
mattn: No need to do `go generate` for build. This is a way to update wmi structs when a bug of wmi2struct is found. For example, type convertion might be wrong, or yet another types must be listed.

https://github.com/mattn/wmi2struct/blob/db7bf4ef341acaef52b9b7bb11c6f5d8e3dee52a/main.go#L32

astj: I want to avoid the case that someone will want to re-generate it for some reason and they don't understand what is necessary for `generat`ing.  So, I still think it's better to be documented wmi2struct (and SQL Server ?) is necessary for this `go generate`.

astj:
Thank you. merged.

mattn:
Thanks!

PR #521, opened 2018-07-23
[mysql] Fix decoding transaction ids from mysql innodb status

itchyny:
This pull request fixes the metric values of `custom.mysql.innodb_transactions.innodb_transactions`. We mistakenly decode transaction id in decimal numbers assuming hex format.

Transaction ids are printed in hex format in version < 5.6.4, but it is now printed in decimal. 
Ref: https://github.com/mysql/mysql-server/commit/3420dc52b68c9afcee0a19ba7c19a73c2fbb2913

### DIFF
--- a/mackerel-plugin-mysql/lib/mysql.go
+++ b/mackerel-plugin-mysql/lib/mysql.go
@@ -1,6 +1,7 @@
 package mpmysql
 
 import (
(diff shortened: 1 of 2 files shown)


itchyny:
Thank you.

PR #522, opened 2018-07-24
[nit] [mysql] prefer strings.HasPrefix / strings.Contains rather than strings.Index

itchyny:
This is just a preference, but strings.HasPrefix should be better than strings.Index(...) == 0.

### DIFF
--- a/mackerel-plugin-mysql/lib/mysql.go
+++ b/mackerel-plugin-mysql/lib/mysql.go
@@ -760,7 +760,7 @@ func parseInnodbStatus(str string, trxIDHexFormat bool, p *map[string]float64) {
 			increaseMap(p, "os_waits", record[8])
 			continue
 		}
-		if strings.HasPrefix(line, "RW-shared spins") && strings.Index(line, ";") > 0 {
+		if strings.HasPrefix(line, "RW-shared spins") && strings.Contains(line, ";") {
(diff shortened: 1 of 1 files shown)


mackerel-plugin-mysql/lib/mysql.go
```diff
@@ -754,33 +754,33 @@ func parseInnodbStatus(str string, trxIDHexFormat bool, p *map[string]float64) {
...
 		// Innodb Semaphores
-		if strings.Index(line, "Mutex spin waits") == 0 {
+		if strings.HasPrefix(line, "Mutex spin waits") {
 			increaseMap(p, "spin_waits", record[3])
 			increaseMap(p, "spin_rounds", record[5])
 			increaseMap(p, "os_waits", record[8])
 			continue
 		}
-		if strings.Index(line, "RW-shared spins") == 0 && strings.Index(line, ";") > 0 {
+		if strings.HasPrefix(line, "RW-shared spins") && strings.Contains(line, ";") {
```
itchyny: Precisely speaking `strings.Contains` is exactly equal to `strings.Index(...) >= 0`, not `> 0` but `strings.Contains` is more readable and does not change the behavior of parsing mysql innodb status output.

itchyny:
Thank you.

PR #528, opened 2018-10-05
Set (default) User-Agent header to HTTP requests

astj:
Previously they're all `Go-http-client/x.y`.
(Probably we should announce this change)

### DIFF
--- a/mackerel-plugin-rack-stats/lib/rack.go
+++ b/mackerel-plugin-rack-stats/lib/rack.go
@@ -62,10 +62,11 @@ func parseBody(r io.Reader, index string) (stats map[string]interface{}, err err
 
 func parseBodyHTTP(uri, port string) (stats map[string]interface{}, err error) {
 	var req *http.Request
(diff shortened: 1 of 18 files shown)


mackerel-plugin-rack-stats/lib/rack.go
```diff
@@ -87,7 +88,12 @@ func parseBodyUnix(path string) (stats map[string]interface{}, err error) {
 	}
 
 	client := &http.Client{Transport: tr}
-	resp, err := client.Get(fmt.Sprintf("http://dummy/%s", strings.TrimLeft(path, "/")))
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://dummy/%s", strings.TrimLeft(path, "/")), nil)
+	if err != nil {
+		return stats, err
+	}
+	req.Header.Set("User-Agent", "mackerel-plugin-rack-status")
```
syou6162: [nit] `mackerel-plugin-rack` would be better?

PR #533, opened 2018-11-27
[postgres] Enable connection without password

kyoshidajp:
My development env of Postgres doesn't need to specify a password. It would be great if without check it.

### DIFF
--- a/mackerel-plugin-postgres/lib/postgres.go
+++ b/mackerel-plugin-postgres/lib/postgres.go
@@ -306,7 +306,12 @@ func (p PostgresPlugin) MetricKeyPrefix() string {
 // FetchMetrics interface for mackerelplugin
 func (p PostgresPlugin) FetchMetrics() (map[string]interface{}, error) {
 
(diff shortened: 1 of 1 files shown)


mackerel-plugin-postgres/lib/postgres.go
```diff
@@ -306,7 +306,12 @@ func (p PostgresPlugin) MetricKeyPrefix() string {
 // FetchMetrics interface for mackerelplugin
 func (p PostgresPlugin) FetchMetrics() (map[string]interface{}, error) {
 
-	db, err := sqlx.Connect("postgres", fmt.Sprintf("user=%s password=%s host=%s port=%s sslmode=%s connect_timeout=%d %s", p.Username, p.Password, p.Host, p.Port, p.SSLmode, p.Timeout, p.Option))
+	cmd := fmt.Sprintf("user=%s host=%s port=%s sslmode=%s connect_timeout=%d %s", p.Username, p.Host, p.Port, p.SSLmode, p.Timeout, p.Option)
+	if p.Password != "" {
+		cmd = fmt.Sprintf("password=%s %s", p.Password, cmd)
+	}
+	db, err := sqlx.Connect("postgres", cmd)
```
kyoshidajp: Actually, always specify a password here, it may not be a problem.

astj:
Thank you. LGTM!

PR #535, opened 2018-12-01
Add redash api-key option

kyoshidajp:
Add `-api-key` option.

Synopsis:

```
mackerel-plugin-redash [-metric-key-prefix=redash] [-timeout=5] \
 [-uri=http://localhost/api/admin/queries/tasks?api_key=hoge] \
 [-api-key=YOUR_REDASH_API_KEY]
```

The value of `-api_key` is used as uri's one if the option is presented or `REDASH_API_KEY` environment variable is set. 

Specially, an admin's API key of Redash is strong. I want to remove it in config file to prevent reading it from other users who are in the host.

### DIFF
--- a/mackerel-plugin-redash/README.md
+++ b/mackerel-plugin-redash/README.md
@@ -6,8 +6,10 @@ redash stats custom metrics plugin for [mackerel-agent](https://github.com/macke
 ## Synopsis
 
 ```shell
-mackerel-plugin-redash [-metric-key-prefix=redash] [-timeout=5] [-uri=http://localhost/api/admin/queries/tasks?api_key=hoge]
+mackerel-plugin-redash [-metric-key-prefix=redash] [-timeout=5] [-uri=http://localhost/api/admin/queries/tasks?api_key=hoge] [-api-key=YOUR_REDASH_API_KEY]
 ```
+
+`REDASH_API_KEY` environment variable is used as `api_key` query in uri if it is set.
 
 ## Example of mackerel-agent.conf
 
--- a/mackerel-plugin-redash/README.md
+++ b/mackerel-plugin-redash/README.md
@@ -6,7 +6,7 @@ redash stats custom metrics plugin for [mackerel-agent](https://github.com/macke
 ## Synopsis
 
 ```shell
-mackerel-plugin-redash [-metric-key-prefix=redash] [-timeout=5] [-uri=http://localhost/api/admin/queries/tasks?api_key=hoge] [-api-key=YOUR_REDASH_API_KEY]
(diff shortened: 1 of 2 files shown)


mackerel-plugin-redash/lib/redash.go
```diff
@@ -93,11 +95,21 @@ func (p RedashPlugin) FetchMetrics() (map[string]interface{}, error) {
 // Do the plugin
 func Do() {
 	optURI := flag.String("uri", "http://localhost/api/admin/queries/tasks?api_key=hoge", "stats URI")
```
astj: With this p-r we should encourage users to use `-api-key` rather than `api_key` parameter in `-uri`.
So can you remove `?api_key=` from this default value?

mackerel-plugin-redash/README.md
```diff
@@ -6,9 +6,11 @@ redash stats custom metrics plugin for [mackerel-agent](https://github.com/macke
 ## Synopsis
 
 ```shell
-mackerel-plugin-redash [-metric-key-prefix=redash] [-timeout=5] [-uri=http://localhost/api/admin/queries/tasks?api_key=hoge]
+mackerel-plugin-redash [-metric-key-prefix=redash] [-timeout=5] [-uri=http://localhost/api/admin/queries/tasks?api_key=hoge] [-api-key=YOUR_REDASH_API_KEY]
```
astj: Same as https://github.com/mackerelio/mackerel-agent-plugins/pull/535/files#diff-3496c6f01c7e15e8d273911dfb989693R97, I'd like to request to remove `?api_key=hoge` from this example

mackerel-plugin-redash/lib/redash.go
```diff
@@ -93,11 +95,21 @@ func (p RedashPlugin) FetchMetrics() (map[string]interface{}, error) {
...
+	apiKey := flag.String("api-key", os.Getenv("REDASH_API_KEY"), "API key")
 	optPrefix := flag.String("metric-key-prefix", "redash", "Metric key prefix")
 	optTimeout := flag.Uint("timeout", 5, "Timeout")
 	optTempfile := flag.String("tempfile", "", "Temp file name")
 	flag.Parse()
 
+	if *apiKey != "" {
```
astj: [optional]
Adding tests around this option is better, but just "better".

astj:
@kyoshidajp Thank you for nice patch! I commented. Please check 🙏

kyoshidajp:
@astj Thank you for your review. I will fix later.

astj:
@kyoshidajp Thank you. (And sorry for late reply!). I'd like to merge this, but I'll modify small change in next p-r.

PR #539, opened 2019-01-10
[redis] Change evicted_keys.Diff to true

lufia:
Sorry, in previous PR #537, I misunderstood `Diff` parameter meaning.
`evicted_keys` should set `Diff` to true because it is counter number.


itchyny:
LGTM

PR #543, opened 2019-02-05
[linux] consider the case where the width of the Netid column is only 5 in the output of ss

Songmu:
The width of ss's Netid column may be a minimum of 5.

https://github.com/sivasankariit/iproute2/blob/1179ab033c31d2c67f406be5bcd5e4c0685855fe/misc/ss.c#L3111-L3117


lufia:
LGTM!

PR #554, opened 2019-06-06
support go modules

astj:


go.mod
```diff
@@ -0,0 +1,36 @@
+module github.com/mackerelio/mackerel-agent-plugins
+
+go 1.12
```
astj: I hope this does not make problems to build with Go 1.11.x ...

lufia: `go` directive describes in Go 1.12 release notes as:

> If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.

https://golang.org/doc/go1.12#modules

astj:
Thanks!

PR #560, opened 2019-07-16
[jvm] fix jinfo command timed out error is not logged

susisu:


astj:
We should also remove other `os.Exit` s during FetchMetrics?  Just returning error and handling the error in FetchMetrics seems more healthy..

astj:
This change itself seems good, as long as CI passes

susisu:
> We should also remove other os.Exit s during FetchMetrics? Just returning error and handling the error in FetchMetrics seems more healthy..

Ah, yes. I changed so f5d644f.

susisu:
Thank you for the review!

PR #571, opened 2019-10-03
[solr] Fix several graph definitions

supercaracal:
Hello,

I found that several metrics values are cumulative in Apache Solr. So they are needed to specify the option `Diff: true`.

### DIFF
--- a/mackerel-plugin-solr/lib/solr.go
+++ b/mackerel-plugin-solr/lib/solr.go
@@ -267,7 +267,8 @@ func (s SolrPlugin) GraphDefinition() map[string]mp.Graphs {
 				path = escapeSlash(path)
 				metricLabel := strings.Title(path)
 				diff := false
-				if key == "requests" {
+				if key == "requests" || key == "errors" || key == "timeouts" ||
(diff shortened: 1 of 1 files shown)


lufia:
@supercaracal  Thank you. The patch looks good.

I'm looking at the documents, [v6.6](https://lucene.apache.org/solr/guide/6_6/performance-statistics-reference.html) and [v7.7](https://lucene.apache.org/solr/guide/7_7/performance-statistics-reference.html).

I think the `timeouts` metric is cumulative too. Can I get your opinion?

supercaracal:
I apologize for my bad fix. I've looked into solr prometheus exporter, then I've found the `timeouts` metric is cumulative.

* https://prometheus.io/docs/concepts/metric_types/#counter
* https://github.com/apache/lucene-solr/blob/1cda4245032d1e0702f9c5e06bafcdfc22bdec60/solr/contrib/prometheus-exporter/conf/solr-exporter-config.xml#L340-L346
* https://qiita.com/mosuka/items/3a5e498c32da820165ef (Japanese)

lufia:
LGTM, Thank you!

PR #575, opened 2019-12-20
add .dependabot/config.yml

lufia:
I added config.yml to be able to enable Dependabot.

ref mackerelio/mackerel-agent#594

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,5 @@
+version: 1
+update_configs:
+- package_manager: go:modules
+  directory: /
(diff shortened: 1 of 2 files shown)


.dependabot/config.yml
```diff
@@ -0,0 +1,5 @@
+version: 1
+update_configs:
+- package_manager: go:modules
+  directory: /
+  update_schedule: daily
```
syou6162: How about `update_schedule: weekly` ? `daily` might be too frequently.

lufia: ok.

PR #583, opened 2020-01-30
[varnish] Added backend_reuse and backend_recycle metrics

cohalz:
If varnish persists communication with the upstream, the values of backend_reuse and backend_recycle also change.

- [Reusing backend connections to increase performance](https://www.fastly.com/blog/reusing-backend-connections-increase-performance)
  - > What you're looking for is for backend_reuse to not be zero, and ideally to be an order of magnitude, or more, higher than backend_conn. 
- [Enormous amount of connections stuck in CLOSE_WAIT state with Varnish](https://stackoverflow.com/questions/27247756/enormous-amount-of-connections-stuck-in-close-wait-state-with-varnish)
  - > And finally, check the output of varnishstat -1 |grep backend to see if varnish is able to reuse backend connections (backend_reuse) and if it has noticed that they are closed (backend_toolate). The values should be so backend_reuse + backend_toolate ≅ backend_recycle.


lufia:
I saw also https://varnish-cache.org/docs/trunk/reference/varnish-counters.html

PR #645, opened 2020-06-24
[plugin-accesslog] allow fields that cannot be parsed but unused by the plugin

### LINKED ISSUE #604: [accesslog] Handle multiple values in apptime
When using the log format `apptime:$upstream_response_time`, the response tiem 
can be multiple values on retry. This results in failure to parse the log line. Maybe only the last value should be used (or summation?).

`ltsv: cannot unmarshal number 0.000, 2.900 into Go value of type float64\n2020/03/24 12:00:00 failed to parse ltsvl`

> Times of several responses are separated by commas and colons like addresses in the $upstream_addr variable.
http://nginx.org/en/docs/http/ngx_http_upstream_module.html

susisu:
Fixes #604

Previously, if there are fields that cannot be parsed in a log, the plugin ignores the whole log line, even if those fields are not used by the plugin.
This PR enables `Loose` flag of the parsers to allow logs to contain fields that cannot be parsed.
The fields used by the plugin such as `Request` and `Status` are still validated to be backward compatible.


mackerel-plugin-accesslog/lib/accesslog.go
```diff
@@ -150,14 +155,25 @@ func (p *AccesslogPlugin) FetchMetrics() (map[string]float64, error) {
...
+			p.parser, l, err = parsers.GuessParser(line)
 		} else {
 			l, err = p.parser.Parse(line)
 		}
 		if err != nil {
 			log.Println(err)
 			continue
 		}
+
+		// ignore invalid logs
```
susisu, line 167: Previously these validations were done by the parsers:

- https://github.com/Songmu/axslogparser/blob/5a6279a7b3d05d0edf047c8538f37eae38aaa1f5/axslogparser.go#L42-L44
- https://github.com/Songmu/axslogparser/blob/5a6279a7b3d05d0edf047c8538f37eae38aaa1f5/apache.go#L60-L64

astj: I agree with you that we need validation for `l.Status`.
Can we omit validation for `l.Request` since we don't depend on actual value of `l.Request`?
I don't think this backward-compatibility is not important (but also, there's no strong demand to remove this validation so we can keep it too, imo)

susisu: I also agree that this validation can be safely removed, hopefully.
But it is still a "breaking change" and I'd like to keep the previous behavior, at least in this PR.

susisu: I thought about this last night, and noticed that this is almost the same as removing validation on `l.AppTime`.
Now I'm going to remove this validation.

susisu:
Thank you!

PR #654, opened 2020-07-20
[plugin-mongodb] add test.sh

lufia:
I added basic tests for mackerel-plugin-mongodb.

ref #629


mackerel-plugin-mongodb/test.sh
```diff
@@ -0,0 +1,35 @@
...
+prog=$(basename $0)
+if ! [[ -S /var/run/docker.sock ]]
+then
+	echo "$prog: there are no running docker" >&2
+	exit 2
+fi
+
+cd $(dirname $0)
+plugin=$(basename $(pwd))
+if ! which -s $plugin
```
astj, line 12: this requires target plugin (in this case `mackerel-plugin-mongodb`) to be installed within `$PATH`, but I'm happy if it works with `$0/mackerel-plugin-mongodb`

mackerel-plugin-mongodb/test.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
astj, line 1: This file has shebang but its permission is `0644`. Should we set this to `0755` ?

mackerel-plugin-mongodb/test.sh
```diff
@@ -0,0 +1,35 @@
...
+docker run -d \
+	--name test-$plugin \
+	-p $port:$port \
+	-e MONGO_INITDB_ROOT_USERNAME=$user \
+	-e MONGO_INITDB_ROOT_PASSWORD=$password \
+	mongo:3-xenial
+trap 'docker stop test-$plugin; docker rm test-$plugin; exit' EXIT
+sleep 10
+
+if $plugin -port $port -username=$user -password $password >/dev/null 2>&1
```
astj, line 30: Is there any reason to redirect stdout/err to /dev/null ?
(One might want to check what plugin says, like some warnings)

mackerel-plugin-mongodb/test.sh
lufia: Originally, I've wanted to run the plugin that was installed in somewhere. For example, if in development, the plugin might there in current directory. Or if in integration test, the plugin in might be installed in **/usr/bin** with the package manager.

So I'm going to add the plugin directory to front of the *PATH*. How do you feel about that?

mackerel-plugin-mongodb/test.sh
lufia: There is no strong reason. I thought outputs of the all plugins could be felt chatty only when run multiple tests.

I will remove this redirect.

mackerel-plugin-mongodb/test.sh
astj: Looks good to me!

PR #766, opened 2021-06-03
[plugin-aws-cloudfront]Replace label name of graph

wafuwafu13:
Replace label name of graph like this.
https://github.com/mackerelio/mackerel-agent-plugins/blob/df9c8643da21869d54cf8625dd7523d88b1c4e20/mackerel-plugin-aws-dynamodb/lib/aws-dynamo-db.go#L315


astj (approved):

astj:
Thank you for p-r and sorry for late response. LGTM!

PR #770, opened 2021-06-08
[plugin-mysql] Ignores non-numerical values

### LINKED ISSUE #769: [mysql] cannot collect metrics after upgrading v0.64.1
We are receiving reports that, after upgrading to v0.64.1, mackerel-plugin-mysql cannot collect metrics and outputs the following log.

```
saveValues: json: unsupported value: +Inf
```

We are now investigating the problem.
If you encountered this issue, please downgrade to v0.64.0 until we release a fixed version.

lufia:
Closes #769

### DIFF
--- a/mackerel-plugin-mysql/UPDATE.md
+++ b/mackerel-plugin-mysql/UPDATE.md
@@ -1,0 +1,11 @@
+# How to check the plugin
+
+## Requirements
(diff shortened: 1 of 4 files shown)


yseto (approved):
LGTM!

PR #775, opened 2021-06-17
[ci][plugin-redis] check metrics

lufia:
I added graphite-metric-test to check metric values with rules. It detects missing metrics, unexpected metrics or out of range metrics.

ref https://pkg.go.dev/github.com/lufia/graphitemetrictest/cmd/graphite-metric-test


yseto (approved):
LGTM!

PR #779, opened 2021-06-25
[plugin-redis] avoid to store +Inf

lufia:
You would be able to look **"saveValues:  json: unsupported value: +Inf"** error  in the result of GitHub Action's output of the first commit of this PR; [756ca59](https://github.com/mackerelio/mackerel-agent-plugins/pull/779/checks?check_run_id=2915016825)
Therefore I fixed it.

see #770


yseto (approved):
LGTM

PR #784, opened 2021-07-02
Replacing the version update process

yseto:
We are using [Mackerel::ReleaseUtils](https://metacpan.org/pod/Mackerel::ReleaseUtils) in the version update process and want to replace it.
In this pull request, I'm writing the code to try the replacement.

This GitHub Action does not actually work, but runs in "dry-run" mode.


lufia (approved):

yseto:
Thanks!

PR #791, opened 2021-07-18
[fluentd] add workers option.

sugy:
add workers option of fluentd plugin.
workers option supports fetching metrics for multi process workers.

refs.

- [Multi Process Workers - Fluentd]( https://docs.fluentd.org/deployment/multi-process-workers )
- [monitor_agent - Fluentd]( https://docs.fluentd.org/input/monitor_agent#multi-process-environment )

### DIFF
--- a/mackerel-plugin-fluentd/lib/fluentd.go
+++ b/mackerel-plugin-fluentd/lib/fluentd.go
@@ -173,7 +173,10 @@ func (f *FluentdPlugin) fetchFluentdMetrics(host string, port int) (map[string]i
 
 // FetchMetrics interface for mackerelplugin
 func (f FluentdPlugin) FetchMetrics() (map[string]interface{}, error) {
-	port, _ := strconv.Atoi(f.Port)
+	port, err := strconv.Atoi(f.Port)
+	if err != nil {
+		return nil, err
+	}
 	if f.Workers > 1 {
 		metrics := make(map[string]interface{})
 		for workerNumber := 0; workerNumber < int(f.Workers); workerNumber++ {
--- a/mackerel-plugin-fluentd/lib/fluentd.go
+++ b/mackerel-plugin-fluentd/lib/fluentd.go
@@ -302,6 +302,9 @@ func (f FluentdPlugin) GraphDefinition() map[string]mp.Graphs {
 	labelPrefix := strings.Title(f.Prefix)
 	graphs := make(map[string]mp.Graphs, len(defaultGraphs))
 	for key, g := range defaultGraphs {
+		if f.Workers > 1 {
+			key = metricName(key, "#")
+		}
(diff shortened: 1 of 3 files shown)


mackerel-plugin-fluentd/lib/fluentd.go
```diff
@@ -298,6 +335,7 @@ func Do() {
 	prefix := flag.String("metric-key-prefix", "fluentd", "Metric key prefix")
 	tempFile := flag.String("tempfile", "", "Temp file name")
 	extendedMetricNames := flag.String("extended_metrics", "", "extended metric names joind with ',' or 'all' (fluentd >= v1.6.0)")
+	workers := flag.Uint("workers", 1, "specifying the number of Fluentd's multi-process workers")
```
lufia, line 338: [ask] Do you think there is a possibility that users of the plugin would like to use the metric name like `custom.fluentd.worker1.#.*` even if `-workers=1`?

If so, I think it is better to set default value to `0`. If not, I think `1` is good value.

lufia (commented):
I have a question, commented.

Please add new option to mackerel-plugin-fluentd/README.md.

mackerel-plugin-fluentd/lib/fluentd.go
sugy: Thanks for your review.

I think it's unlikely, because that users with workers=1 will not be aware of workers.
The default number of workers in fluentd is one.

ref. https://docs.fluentd.org/deployment/multi-process-workers#multi-process-workers-and-plugins

> By default, one instance of `fluentd` launches a supervisor and a worker. A worker consists of input/filter/output plugins.

Also, if users want to get the metrics of a specific worker, they can use the `-metric-key-prefix` option and the `-port` option.

mackerel-plugin-fluentd/lib/fluentd.go
```diff
@@ -168,6 +171,35 @@ func (f FluentdPlugin) FetchMetrics() (map[string]interface{}, error) {
 	return f.parseStats(body)
 }
 
+// FetchMetrics interface for mackerelplugin
+func (f FluentdPlugin) FetchMetrics() (map[string]interface{}, error) {
+	port, _ := strconv.Atoi(f.Port)
```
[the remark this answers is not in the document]
sugy: Fixed. 49cb3159dac6d30a2251d5554af487945fd74961

mackerel-plugin-fluentd/lib/fluentd.go
lufia: I see, it looks good.

lufia:
@sugy Thank you!

lufia (approved):

sugy:
@lufia Thanks!

PR #795, opened 2021-07-29
[accesslog] use Seek to skip the log

lufia:
It is not good to use ioutil.ReadAll to skip all logs.

ioutil.ReadAll reads all bytes into memory, so if the access log is large then the plugin might be OOM crash.


ne-sachirou (approved):
🙆‍♀️

ne-sachirou (approved):

PR #802, opened 2021-08-11
added "pending log flushes"

yseto:
There is a difference in the values that can be obtained depending on the mysql version.

Followed the previous implementation to retrieve values, but changed the labels to avoid confusion because different values are organized with similar names.

Also, the values that were failing to be retrieved are now retrieved.


susisu (approved):
LGTM!

PR #807, opened 2021-09-01
[mysql] adapt for mysql5.7's InnoDB SEMAPHORES metrics

do-su-0805:
## why

Starting with mysql5.7, the metric for "RW-sx" is displayed. I want to observe this.

## how

- Added according to examples such as "RW-shared".

## check

- Correct the test and make sure it passes.

### unchecked

- I have not confirmed the behavior in MySQL8.0.

## FYI

- Starting with mysql5.7.8, The "Mutex spin waits" line is no longer displayed.
  - https://github.com/mysql/mysql-server/commit/3574d439c1cf7ad39a5077d5d3c7ff5ad835a882
  - https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-8.html

### DIFF
--- a/mackerel-plugin-mysql/lib/mysql.go
+++ b/mackerel-plugin-mysql/lib/mysql.go
@@ -834,6 +834,12 @@ func parseInnodbStatus(str string, trxIDHexFormat bool, p map[string]float64) {
 			increaseMap(p, "os_waits", record[7])
 			continue
 		}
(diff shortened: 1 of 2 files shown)


do-su-0805:
On the other hand, I'm not sure why I didn't look at the spin_rounds of RW-XX.

yseto (approved):
@do-su-0805 thank you!

PR #809, opened 2021-09-27
[plugin-redis] migrate redis client library to redigo

pyto86pri:
The redis client library `fzzy/radix` seems to be removed, so I migrated the library to `gomodule/redigo`.

### DIFF
--- a/mackerel-plugin-redis/lib/redis.go
+++ b/mackerel-plugin-redis/lib/redis.go
@@ -3,6 +3,7 @@ package mpredis
 import (
 	"flag"
 	"fmt"
+	"net"
 	"os"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


mackerel-plugin-redis/lib/redis.go
```diff
@@ -106,12 +94,12 @@ func (m RedisPlugin) MetricKeyPrefix() string {
 // FetchMetrics interface for mackerelplugin
 func (m RedisPlugin) FetchMetrics() (map[string]interface{}, error) {
 	network := "tcp"
-	target := fmt.Sprintf("%s:%s", m.Host, m.Port)
+	address := fmt.Sprintf("%s:%s", m.Host, m.Port)
```
lufia, line 97: [nit] is better to use `net.JoinHostPort`

mackerel-plugin-redis/lib/redis.go
```diff
@@ -303,13 +286,13 @@ func (m RedisPlugin) GraphDefinition() map[string]mp.Graphs {
...
 	if m.Socket != "" {
-		target = m.Socket
 		network = "unix"
+		address = m.Socket
 	}
 
-	c, err := redis.DialTimeout(network, target, time.Duration(m.Timeout)*time.Second)
+	c, err := redis.Dial(network, address, redis.DialConnectTimeout(time.Duration(m.Timeout)*time.Second))
```
lufia, line 295: [memo] reids package has DialTimeout, but it had been marked as deprecated.

lufia (approved):

mackerel-plugin-redis/lib/redis.go
pyto86pri: Make sense 👍

PR #874, opened 2022-04-21
[plugin-mysql] fix panic when parsing aio stats

lufia:
Amazon Aurora could return aio statistics like below:

```
Pending normal aio reads:, aio writes: [0, 0, 0, 0] ,
```

That statistics is missing values of reads so plugin-mysql panics when parsing aio line.


susisu (approved):
LGTM

PR #894, opened 2022-06-08
[plugin-docker] (Breaking) drop 'File' method support

lufia:
In the past, Docker 1.5.x or earlier didn't have APIs to retrieve metrics so *plugin-docker* was read any metrics from files.
Today, that behavior still be supported if `-method File` option are passed to *plugin-docker* command.

However I think we could stop to support that option because Docker 1.6 had been released in 2015 and it is already marked as EOL.

This PR deletes `-method File` related codes and makes to print a message and exit if `-method File` is passed.

### DIFF
--- a/mackerel-plugin-docker/lib/docker.go
+++ b/mackerel-plugin-docker/lib/docker.go
@@ -1,15 +1,12 @@
 package mpdocker
 
 import (
-	"bytes"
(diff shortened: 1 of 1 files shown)


susisu (approved):
LGTM!

PR #945, opened 2022-10-28
[mongodb] add metric-key-prefix option

tukaelu:
The metric-key-prefix option has been added to mackerel-plugin-mongodb to resolve issues that have been inquired to support.

In line with this, the graph labels now also reflect the prefix.


yseto (approved):
seems good.

PR #961, opened 2022-12-12
Purge nvidia smi

lufia:
We have decided some old and less used plugins would be purge from the official plugin packages.

https://github.com/mackerelio/mackerel-plugin-nvidia-smi


yseto (approved):
seems good.

PR #966, opened 2022-12-15
[plugin-elasticsearch] Fix the test for elasticsearch

lufia:
I saw the test was sometimes failed because elasticsearch container do not become running state even if test script sleeps a minute.


yseto (approved):
seems good

PR #969, opened 2022-12-15
remove xentop on backward compatibility symlink

yseto:


lufia (approved):
note: these lists are not changed by perl script.

PR #972, opened 2022-12-16
split go.mod for plugins that have previously split repositories

yseto:
refers
- https://github.com/mackerelio/mackerel-agent-plugins/pull/961
- https://github.com/mackerelio/mackerel-agent-plugins/pull/962


lufia (approved):

yseto:
Thanks for your review.

PR #976, opened 2022-12-19
accesslog: use `reqtime_microsec` if exists

wafuwafu13:


lufia (approved):

lufia:
:+1: 

Would be better to change unit of "latency" to "seconds"?

**Note**: In Mackerel, Graph defs never change automatically if it is created once.

mackerel-plugin-accesslog/lib/accesslog.go
```diff
@@ -179,7 +180,9 @@ func (p *AccesslogPlugin) FetchMetrics() (map[string]float64, error) {
 		ret[string(fmt.Sprintf("%d", l.Status)[0])+"xx_count"]++
 		ret["total_count"]++
 
-		if l.ReqTime != nil {
+		if l.ReqTimeMicroSec != nil {
+			reqtimes = append(reqtimes, *l.ReqTimeMicroSec*math.Pow(10, -6))
```
lufia, line 184: https://go.dev/play/p/EPsEBCpofCZ

PR #1001, opened 2023-01-23
fix staticcheck

yseto:


mackerel-plugin-solr/lib/solr.go
```diff
@@ -246,7 +246,7 @@ func (s SolrPlugin) GraphDefinition() map[string]mp.Graphs {
 		}
 
 		for _, key := range []string{"indexHeapUsageBytes", "segmentCount", "sizeInBytes"} {
-			metricLabel := strings.Title(key)
+			metricLabel := strings.Title(key) // nolint
```
lufia, line 249: I think this `// nolint` requires a comment why `strings.Title` is kept here.

eg. `// nolint because this label will use in solr plugin's graph-defs, it should keep same representation.`

lufia (approved):

yseto:
Thanks for your review.

PR #1055, opened 2023-04-04
Fix timeout/error labels of mackerel-plugin-twemproxy

kmuto:
In mackerel-plugin-twemproxy, `Server Timeout` and `Server Error` are reversed in the display label and value set.
This patch trys to fix it.


yseto (approved):
seems good.

PR #1072, opened 2023-05-26
[fluentd] add README.md with extended_metrics option

sakamossan:
I found an option that was added in #738 but it is not described in the README.


yseto (approved):
seems good. this was very helpful.

PR #1155, opened 2024-02-28
[mackerel-plugin-windows-server-sessions] Remove the dependency for WMIC command

Arthur1:
WMIC is deprecated as of the 21H1 semi-annual channel release of Windows Server. cf.) https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic

So, I have updated mackerel-plugin-windows-server-sessions to use a Go library that interact directly with WMI instead of WMIC command.

### before

```console
> WMIC PATH Win32_PerfFormattedData_PerfNet_Server GET ServerSessions /FORMAT:CSV

Node,ServerSessions
EC2AMAZ-BAJQD2D,0
```

```console
> cd mackerel-plugin-windows-server-sessions
> go run ./main
windows.server.sessions.EC2AMAZ-BAJQD2D.count   0       1709118187
```

### after

```console
> go run ./main
windows.server.sessions.EC2AMAZ-BAJQD2D.count   0       1709127711
```

### DIFF
--- a/mackerel-plugin-windows-server-sessions/lib/sessions.go
+++ b/mackerel-plugin-windows-server-sessions/lib/sessions.go
@@ -1,3 +1,5 @@
+//go:build windows
+
 package mpwindowsserversessions
 
 import (
--- a/mackerel-plugin-windows-server-sessions/lib/sessions.go
+++ b/mackerel-plugin-windows-server-sessions/lib/sessions.go
@@ -14,6 +14,7 @@ import (
 
 // WindowsServerSessionsPlugin store the name of servers
 type WindowsServerSessionsPlugin struct {
+	LegacyMetricName bool
 }
 
 // cf.) https://learn.microsoft.com/en-us/previous-versions/aa394265(v=vs.85)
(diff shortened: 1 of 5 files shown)


mackerel-plugin-windows-server-sessions/lib/sessions.go
```diff
@@ -1,39 +1,50 @@
...
-	counts := make(map[string]int)
-	for _, record := range records[1:] {
-		name := strings.TrimSpace(record[0])
-		n, err := strconv.Atoi(strings.TrimSpace(record[1]))
-		if err != nil {
-			continue
-		}
-		counts[name] = n
+	counts := []ServerSessionsCount{
+		{Node: node, ServerSessions: dst[0].ServerSessions},
```
Arthur1, line 45: The plugin was originally implemented as if it could retrieve metrics for multiple nodes, but in reality it can only retrieve metrics for one node at most.

mackerel-plugin-windows-server-sessions/lib/sessions.go
```diff
@@ -43,9 +54,11 @@ func (m WindowsServerSessionsPlugin) FetchMetrics() (map[string]interface{}, err
 	if err != nil {
 		return nil, err
 	}
-	stat := make(map[string]interface{})
-	for k, v := range counts {
-		stat["windows.server.sessions."+k+".count"] = uint64(v)
+	stat := make(map[string]interface{}, len(counts))
+	for _, v := range counts {
+		// node name of Windows can contain ".", which is the metric name delimiter on Mackerel.
+		nodeMetricKey := strings.ReplaceAll(v.Node, ".", "_")
```
lufia, line 60: In my opinion, keeping `.` in the metric name was a bug, so this does not break compatibility.

lufia (approved):

mackerel-plugin-windows-server-sessions/lib/sessions.go
lufia: Of course, I think it is ok to add `-legacy` option to keep behavior. But it is not required.

Arthur1: Because there are some hosts with node name including ".", we added `-legacymetricname` option. 280503b

PR #1165, opened 2024-03-28
Allow mackerel-plugin-redis to specify username

mkadokawa-idcf:
The mackerel-plugin-redis plugin allows you to specify username.
If nothing is set, it will be the default user.


Arthur1 (approved):
Awesome. I made an additional commit to update README.

Thank you for your contribution!

PR #1166, opened 2024-03-29
add STRING number support

kmuto:
I found mackerel-plugin-snmp failed to parse a number of STRING type from SNMP device, like `strconv.ParseFloat: parsing "[51 46 49 52]": invalid syntax`.
For example, original value from SNMP device is '3.14', and `51 46 49 52` seems the byte array of '3.14'.
Because it is usually impossible to change a type on the SNMP device side, the receiver has a responsiblity to accept it.

Here is a patch to support STRING number.

- `parseFloat` again for byte array with forcing String conversion by `fmt.Sprintf('%s'` only if the parsing failed first. `resp.Variables[0].Value` may be vary, so original first `fmt.Sprint(resp.Variables[0].Value)` works well, except a byte array of string.
- To test it, extend command is added. It will back STRING 3.14. `.1.3.6.1.4.1.8072.1.3.2.3.1.2.4.101.99.104.111 = STRING: 3.14`

Result of test.sh
Before:
```
2024/03/29 08:47:18 strconv.ParseFloat: parsing "[51 46 49 52]": invalid syntax
graphite-metric-test: <stdin>: rule snmp.echo[>0] is not matched any metrics
test-mackerel-plugin-snmp
test-mackerel-plugin-snmp
```

After:
```
test-mackerel-plugin-snmp
test-mackerel-plugin-snmp
```

### DIFF
--- a/mackerel-plugin-snmp/lib/snmp.go
+++ b/mackerel-plugin-snmp/lib/snmp.go
@@ -52,6 +52,12 @@ func (m SNMPPlugin) FetchMetrics() (map[string]interface{}, error) {
 
 		ret, err := strconv.ParseFloat(fmt.Sprint(resp.Variables[0].Value), 64)
 		if err != nil {
+			// NOTE: Cannot assume strconv.ParseFloat("%s", resp.Variables[0].Value)
+			// first, as resp.Variables[0].Value may be an int class, etc.
+			// Normally, an object values such as INTEGER or Counter are
+			// successfully accepted in the above conversions.
+			// However, STRING object values are passed as byte arrays, so the above
+			// conversion will result in an error.
 			ret, err = strconv.ParseFloat(fmt.Sprintf("%s", resp.Variables[0].Value), 64)
 			if err != nil {
 				log.Println(err)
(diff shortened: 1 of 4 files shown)


mackerel-plugin-snmp/lib/snmp.go
```diff
@@ -52,8 +52,11 @@ func (m SNMPPlugin) FetchMetrics() (map[string]interface{}, error) {
 
 		ret, err := strconv.ParseFloat(fmt.Sprint(resp.Variables[0].Value), 64)
 		if err != nil {
-			log.Println(err)
-			continue
+			ret, err = strconv.ParseFloat(fmt.Sprintf("%s", resp.Variables[0].Value), 64)
+			if err != nil {
+				log.Println(err)
+				continue
+			}
```
lufia, line 59: IMO, It's hard to understand immediately why second sentence is needed.

Can you add a comment, or add an explain text in commit message?

kmuto: Sure, I added.

lufia (approved):

PR #1214, opened 2024-08-07
[mackerel-plugin-snmp] support 32bit counter overflow.

yseto:
I did it.

It is useful in cases such as ifInOctets `1.3.6.1.2.1.2.2.1.10.(ifIndex)`.

### DIFF
--- a/mackerel-plugin-snmp/README.md
+++ b/mackerel-plugin-snmp/README.md
@@ -7,9 +7,18 @@ SNMP V2c custom metrics plugin for mackerel.io agent.
 can specify multiple metric-definitions in the form of `OID:NAME[:DIFF?][:STACK?]` args.
 
 ```shell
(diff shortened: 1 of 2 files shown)


mackerel-plugin-snmp/README.md
```diff
@@ -7,10 +7,19 @@ SNMP V2c custom metrics plugin for mackerel.io agent.
 can specify multiple metric-definitions in the form of `OID:NAME[:DIFF?][:STACK?]` args.
```
yohfee, line 7: btw, Is it ok not to change this usage?

yohfee (approved):
Looks good!

mackerel-plugin-snmp/README.md
yseto: Yes. this is new part of command arguments. when Not specified, No change in behavior.

PR #1247, opened 2025-01-23
Fix CI build

ne-sachirou:
The CI for the master branch is failing due to the absence of the `build-essential:native` package, as seen in [this run](https://github.com/mackerelio/mackerel-agent-plugins/actions/runs/12500547540/job/34876951429). It appears that the cause is the change to Ubuntu 24.04, as the runner is specified to use ubuntu-latest in GHA.
This pull request specifies the runner to use ubuntu-22.04.


yseto (approved):
seems good.

PR #1254, opened 2025-02-17
support memory peak value which is introduced in PHP 8.4 on mackerel-plugin-php-fpm

kmuto:
PHP 8.4 introduces new field `memory peak` for php-fpm status page.
https://github.com/php/php-src/blob/0454901ed1518cfb42bc77ac91728591b58c3e6e/UPGRADING#L339
The memory usage peak since FPM started.

This PR subjects to add new metric and graph def to support it on mackerel-plugin-php-fpm plugin.

Since memory peak does not fit into the existing PHP-FPM plugin graph definition, I have added a new one.
Max Active Processes is a similar graph that keeps the peaks since PHP-FPM was started, so I used that as a reference. However, I changed the units to bytes.

To avoid creating meaningless graphs in versions older than PHP 8.4, I have coded to not send out a metric when the value of a field cannot be retrieved (=0).
This would mean that the graph would not be generated even when it was really 0 (e.g. after startup). However, rather than introduce the complexity of checking if it is nil or not, I believe it is no problem since a positive value will be entered as soon as PHP-FPM is called.

~PHP8.3 `status?json`
```
{"pool":"www","process manager":"dynamic","start time":1739778681,"start since":4817,"accepted conn":0,"listen queue":0,"max listen queue":0,"listen queue len":0,"idle processes":2,"active processes":0,"total processes":2,"max active processes":0,"max children reached":0,"slow requests":0}
```

PHP8.4 `status?json`
```
{"pool":"www","process manager":"dynamic","start time":1739783779,"start since":487,"accepted conn":1,"listen queue":0,"max listen queue":0,"listen queue len":0,"idle processes":2,"active processes":0,"total processes":2,"max active processes":1,"max children reached":0,"slow requests":0,"memory peak":2097152}
```

PHP8.3
image

PHP8.4
image

Definition
image

### DIFF
--- a/mackerel-plugin-php-fpm/lib/php-fpm.go
+++ b/mackerel-plugin-php-fpm/lib/php-fpm.go
@@ -130,6 +130,7 @@ type PhpFpmStatus struct {
 	MaxActiveProcesses uint64 `json:"max active processes"`
 	MaxChildrenReached uint64 `json:"max children reached"`
 	SlowRequests       uint64 `json:"slow requests"`
+	MemoryPeak         uint64 `json:"memory peak"`
 }
 
 // MetricKeyPrefix interface for PluginWithPrefix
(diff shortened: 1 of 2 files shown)


yseto (approved):
seems good.

PR #1277, opened 2025-05-07
use Go 1.24

yseto:
we did.


ne-sachirou (approved):
🙆‍♀️

yseto:
Thanks for your review.

PR #1278, opened 2025-05-07
git commit, version from runtime/debug

yseto:
- Go 1.24 or above, can read package version from [BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo)
- also can use vcs.revision
- therefore we can remove build option `-X main.gitcommit`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,12 +82,14 @@ deb-v2-x86:
 	$(MAKE) build/mackerel-plugin GOOS=linux GOARCH=amd64
 	cp build/mackerel-plugin packaging/deb-v2/debian/
 	cd packaging/deb-v2 && debuild --no-tgz-check -rfakeroot -uc -us
+	git clean -f -d ./packaging
 
 .PHONY: deb-v2-arm
 deb-v2-arm:
 	$(MAKE) build/mackerel-plugin GOOS=linux GOARCH=arm64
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -83,12 +82,14 @@ deb-v2-x86:
 	$(MAKE) build/mackerel-plugin GOOS=linux GOARCH=amd64
 	cp build/mackerel-plugin packaging/deb-v2/debian/
 	cd packaging/deb-v2 && debuild --no-tgz-check -rfakeroot -uc -us
+	git clean -f -d ./packaging
```
lufia, line 85: I understand this *git clean* is there to keep **debug.BuildInfo** as clean.

If so, I prefer that to put the line "before" build because the build process will be able to exit unexpectedly.

lufia (approved):

Makefile
yseto: I did. 
https://github.com/mackerelio/mackerel-agent-plugins/pull/1278/commits/02bc4acf36f7d5df72de50a8b1b15b7ad8808854

also added same clean process on before tar builds
https://github.com/mackerelio/mackerel-agent-plugins/pull/1278/commits/3cd171a91f1fbabc03742788a8b5513d8a7233bf

yseto:
Thanks for your review.

PR #1279, opened 2025-05-07
read VERSION from git

yseto:
- read VERSION from git.
  - therefore, No longer needed rewrite Makefile on every releases.
- also stopped rename tarball on every releases.


Makefile
```diff
@@ -1,4 +1,5 @@
-VERSION = 0.88.1
+# This VERSION variable indicates the latest tag.
+VERSION = $(subst v,,$(shell git describe --abbrev=0 --tags))
```
lufia, line 2: note: We guard to not to be able to ship added commits after tagged commit by our workflow (not GitHub Workflows).

lufia (approved):

PR #1306, opened 2025-08-01
add -port option to mackerel-plugin-snmp plugin

kmuto:
The mackerel-plugin-snmp is fixed to use port 161, but depending on the test or environment, user may want to change it to another (in particular, non-privileged) port.
Adding the configuration was easy to do, so here is a patch.


mackerel-plugin-snmp/lib/snmp.go
```diff
@@ -33,6 +34,7 @@ func (m SNMPPlugin) FetchMetrics() (map[string]interface{}, error) {
 	stat := make(map[string]interface{})
 
 	gosnmp.Default.Target = m.Host
+	gosnmp.Default.Port = uint16(m.Port)
```
yseto, line 37:
```suggestion
	gosnmp.Default.Port = m.Port
```

yseto (approved):

PR #1323, opened 2025-09-08
[nit]Revert "Release version 0.89.0"

yseto:
Reverts mackerelio/mackerel-agent-plugins#1322


yseto:
Thanks for your review. :bow:

PR #1327, opened 2025-09-12
fix dependabot config

yseto:
Clarification of files to be updated.


yseto:
Thanks for your review.

PR #1463, opened 2026-04-10
Use Go 1.26 for builds, supports Go 1.25 and above.

ne-sachirou:
- Build with Go 1.26.
- Testing in Go 1.25 and above.
- `go fix` to new syntax.


yseto (approved):
seems good.

